PR #11, opened 2019-05-31
Packet: Fix typos and improve wording for hardware reservation docs

rata:
High level description of the change.

This address all the comments after https://github.com/kinvolk/lokomotive-kubernetes/pull/10 was merged.

The only comment still opened is not addressed (as I think we may want to think a little bit about it before doing it). If we change that too, I will of course add another change for it.


johananl:
Thanks, this is better. I just rendered the docs though and descriptions for the new vars don't follow the convention of the existing vars:

image

Do you think we can make the description more concise and include extra info below the table?

rata:
@johananl thanks for the review, and sorry for taking so long! I've made the description more concise while linked to the packet doc bellow as you recommended. PTAL :)

I attach an image of how it looks now in the docs (but note that it seems large, but it probably is my small laptop monitor as the same happens with taints, that was there before and is not long).

asd

invidian (approved):

rata:
I'd prefer to merge this after @johananl review, as this PR is to address his comments :)

rata:
@johananl friendly ping?

rata:
@johananl friendly ping number 2 :)

invidian:
image

surajssd (approved):
LGTM

surajssd:
I think this can be merged now, if there are any issues with this I think folks can raise it later? WDYT @rata

rata:
Okay, let's merge :)

PR #14, opened 2019-06-03
Remove Github issue/PR templates

johananl:
Removing these Typhoon-originated templates since we aren't using them.


invidian (approved):
:see_no_evil:

PR #18, opened 2019-06-17
azure: make workers compatible with OpenEBS

invidian:
We have the same patches for worker pools for other providers, they are required to run OpenEBS.


johananl (approved):
LGTM

PR #21, opened 2019-06-18
Raid setup improvements

rata:
This PR includes 3 commits:
 * One to fix MD arrays renames that sometimes happen on reboot
 * To escape dollar sign in template correctly
 * And the interesting one: to force RAID 0 creation even if 1 disk is available.

I paste the commit message form the last one, as it is the most interesting one. Please have a look at the rename fix commit message too, but I think that is mostly straight forward.

Here c&p the commit message for using RAID 0 even with 1 disk:

    Even if it is only one spare disk, it makes sense to create a RAID 0
    array because the user might want to use that local disk, it adds no
    performance penalty, and it is uniform with other setups where more
    disks are involved (from the maintenance point of view).
    
    The other option would be to provide a totally different way to access
    the disk (without RAID), but that means more complicated provisioning
    code for a use case we are not fans of.
    
    Let's extend the RAID 0 to these cases, that also doesn't expand the
    test-matrix, and if there is any problem with the approach we can
    re-evaluate later.

### DIFF
--- a/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
@@ -73,11 +73,17 @@ storage:
             )
             local count=$(echo "$disks" | wc -w)
 
-            # Exit if we have less than 2 RAID-eligible disks.
-            [ $count -lt 2 ] && return 0
+            # Exit if we don't have any disks to create an array
+            [ $count -lt 1 ] && return 0
 
             # Create, format and mount array.
-            mdadm --create --verbose /dev/md/node-local-storage --level=0 --raid-devices="$${count}" $${disks}
+            local extra_opts=""
+            if [ $count -lt 2 ]; then
+              # Force array creation even with one disk
+              extra_opts="--force"
+            fi
+
+            mdadm --create $$extra_opts --verbose /dev/md/node-local-storage --level=0 --raid-devices="$${count}" $${disks}
             cat /proc/mdstat
             mkfs.ext4 /dev/md/node-local-storage
           }
--- a/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
@@ -83,7 +83,7 @@ storage:
               extra_opts="--force"
             fi
 
-            mdadm --create /dev/md/node-local-storage $$extra_opts --verbose --level=0 --raid-devices="$${count}" $${disks}
+            mdadm --create /dev/md/node-local-storage --homehost=any $$extra_opts --verbose --name=node-local-storage --level=0 --raid-devices="$${count}" $${disks}
             cat /proc/mdstat
             mkfs.ext4 /dev/md/node-local-storage
           }
--- a/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
@@ -83,7 +83,7 @@ storage:
               extra_opts="--force"
             fi
 
-            mdadm --create $$extra_opts --verbose /dev/md/node-local-storage --level=0 --raid-devices="$${count}" $${disks}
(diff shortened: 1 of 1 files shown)


packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
```diff
@@ -73,11 +73,17 @@ storage:
             )
             local count=$(echo "$disks" | wc -w)
 
-            # Exit if we have less than 2 RAID-eligible disks.
-            [ $count -lt 2 ] && return 0
-
+            # Exit if we don't have any disks to create an array
+            [ $count -lt 1 ] && return 0
+
```
invidian: Not needed whitespace.

invidian (changes requested):
I have some concerns regarding `--hostname=any`, since it indicates that the root cause of the problem is that the hostname is changing at some point, which is unexpected. Could it be that this happens when you boot the installler over iPXE? Maybe we should set the hostname before creating a RAID array then? I feel that we lack full understanding of what's going on here.

Other than that, proposal with single RAID 0 array make sense IMO, though I'm not sure if it simplifies the setups. BTW, according to some benchmarks it may even have some performance benefits (when you use RAID-controller cache when writing).

Also some small nit regarding whitespace.

rata:
@invidian yes, the rename happens after booting over iPXE and reboot. The hostname seems to be set when booting over iPXE, but maybe `gethostname(2)` changes anyways (setting the hostname has it's pitfalls). I can try to check that, if you want. But, if gethostbyname(2) changes, I don't want to change the bootstrap process to change it :-)

Anyways, we don't want the array to be renamed in any case and therefore I think the setting makes sense either way.

If something is wrong, we surely want to fix it (and we can create an issue to investigate the root cause of this), but I don't see why not use this option that will reduce the impact of any issues of these kind. The root cause can (and should) be investigated, but we should also reduce the impact of bugs whenever possible. This option in particular can change from downtime till manual intervention (to manually mount the renamed array, for example) to no downtime at all.

Even if we fix the root cause, I don't see why not use this option to be more resilient in the future.

Am I missing something?  What do you think?

invidian (commented):
>Also, while we are there, the command was changed to use the documented
order of parameters:
 mdadm [mode] <raiddevice> [options] <component-devices>

This could go to separated commit actually, so later addition is obvious while reviewing.

invidian (commented):
Except my last review comment, LGTM, even though I don't like `--hostname=any` solution, as it seems hacky to me, as I think the issue lays in some hostname inconsistency during booting/installation. Maybe there is a race between setting hostname and mdadm initialization?

Anyway, if someone else approves it, I'm fine with merging it.

rata:
> Except my last review comment

Split the commit for this

kosyfrances (approved):
LGTM

rata:
I can confirm that:
 * The output of `hostname` and `hostname -f` is the worker name, by running it just before the mdadm command
 * Setting `--homehost=<worker-name>` and adding `echo HOMEHOST <system> | tee -a /etc/mdadm.conf` just above this line: https://github.com/kinvolk/lokomotive-kubernetes/blob/master/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl#L133 doesn't fix it either. The rename is still done.
 * After a quick look at mdadm changelog, it doesn't seem to be anythig obvious either (but, I might be missing something, I didn't look closely).
 * The mdadm version seems old, however: 
   * `$ mdadm --version`
   * `mdadm - v3.4 - 28th January 2016`


Will open an internal issue to track the root cause, but I think this option should stay even when we fix it (as I said before). Bugs may happen, and this just reduces the impact of several kind of issues. And we *do* know that this MD array won't be used in any other server, so really no need to worry about the array used on another server.

rata:
The root cause seems to be the one described here: https://github.com/coreos/bugs/issues/2104 (thanks @dongsupark) and the fix we did here is the one suggested there too (by chance :))

This homehost option has even been added to ignition here: https://github.com/coreos/ignition/pull/425/files

So, this seems the right fix. Will open an issue to improve RAID setup to use ignition directly (with this very same fix).

PR #22, opened 2019-06-18
Packet: Add label with ip to nodes

miry:
We require for automatization have information about public ip address of an instance.

### DIFF
--- a/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
@@ -101,7 +101,7 @@ systemd:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/master \
           --node-labels=node-role.kubernetes.io/controller="true" \
-          --node-labels=kubernetes.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
+          --node-labels=lokomotive.alpha.kinvolk.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
           --pod-manifest-path=/etc/kubernetes/manifests \
           --read-only-port=0 \
           --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
--- a/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
@@ -101,6 +101,7 @@ systemd:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/master \
           --node-labels=node-role.kubernetes.io/controller="true" \
+          --node-labels=kubernetes.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
           --pod-manifest-path=/etc/kubernetes/manifests \
           --read-only-port=0 \
           --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
--- a/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
@@ -90,7 +90,7 @@ systemd:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/node \
           --node-labels=${worker_labels} \
-          --node-labels=kubernetes.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
+          --node-labels=lokomotive.alpha.kinvolk.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
           --pod-manifest-path=/etc/kubernetes/manifests \
           --read-only-port=0 \
           --register-with-taints=${taints} \
--- a/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
+++ b/packet/flatcar-linux/kubernetes/workers/cl/worker.yaml.tmpl
@@ -90,6 +90,7 @@ systemd:
           --network-plugin=cni \
           --node-labels=node-role.kubernetes.io/node \
           --node-labels=${worker_labels} \
+          --node-labels=kubernetes.io/public-ipv4=$${COREOS_PACKET_IPV4_PUBLIC_0} \
           --pod-manifest-path=/etc/kubernetes/manifests \
           --read-only-port=0 \
           --register-with-taints=${taints} \


rata:
@miry thank you for the patch!

Can you please elaborate on the use case to understand why you need this as a label and not to be consumed from some other place? I'd like to understand the use case in as much detail as possible.

Also, we need to consider:
 * I see this is only added to the controller, I guess only the controller IP is needed for that use case? Do we need to do this on workers too? Please go in as much detail as possible to explain your use case :)

 * Do you need the IP or will something else will work too? For example, is the hostname or some DNS record pointing to the IP enough? If all nodes get a DNS record using their hostname, the hostname might be enough

 * What are other tools doing (kops, for example)? Are they using other name for the key? Maybe there is a well-known label key for this

 * Do we want this to be a label or an annotation? (if it's going to be used by a machine, this might fit better as an annotation)

 * Do we want to hardcode this on every cluster? Maybe this can just be added in terraform. Not sure this is better, but something like this can be added to the module:
 ```
   labels = "rata=$${COREOS_PACKET_IPV4_PUBLIC_0}"
```
 * Not sure if there are better ways so this is set by the user other than the above hack, but I'm definitely interested to explore them. Do you know if there is any other? That way we don't impose any particular label (in this case `kubernetes.io/public-ipv4`)

**Comment by miry** [comment]
> I see this is only added to the controller, I guess only the controller IP is needed for that use case? Do we need to do this on workers too? Please go in as much detail as possible to explain your use case :)

I forgot to put changes for nodes. Updated pr.

**Comment by miry** [comment]
> Do you need the IP or will something else will work too? For example, is the hostname or some DNS record pointing to the IP enough? If all nodes get a DNS record using their hostname, the hostname might be enough.

Interesting proposal, but mostly we need to know public IPs, to automate access to the nodes.
DNS record using their hostname would work, but It would require extra terraform resources to keep it update.

> Do we want to hardcode this on every cluster? Maybe this can just be added in terraform. Not sure this is better, but something like this can be added to the module: `labels = "rata=$${COREOS_PACKET_IPV4_PUBLIC_0}"`

It is a nice solution. I see only one problem is duplication. We use 6-10 modules. And this line would be duplication.

I decided to use label `kubernetes.io/public-ipv4` similar to standard label: `kubernetes.io/hostname` and flannel annotation `flannel.alpha.coreos.com/public-ip`. Open for any suggestions.

Another way to sync metadata with packet and put to annotations.

**Comment by invidian** [comment]
I think the correct way of doing this would be to use https://github.com/packethost/packet-ccm, which should automatically append those labels to all nodes.

**Comment by rata** [comment]
@invidian do you know which labels it adds? So we can use the same.

From a quick look at the code, I dind't see it (it may be logic from a more generic framework to implement CCM that adds it, didn't check).

**Comment by rata** [comment]
> > Do you need the IP or will something else will work too? For example, is the hostname or some DNS record pointing to the IP enough? If all nodes get a DNS record using their hostname, the hostname might be enough.
> 
> Interesting proposal, but mostly we need to know public IPs, to automate access to the nodes.
> DNS record using their hostname would work, but It would require extra terraform resources to keep it update.

So, can you elaborate on your use case, so we know all the possible options too? And evaluate if we maybe want this as a annotation rather than a label?

Btw, I think there might be easy ways to automate the DNS creation from terraform using this module output, for example.

> > Do we want to hardcode this on every cluster? Maybe this can just be added in terraform. Not sure this is better, but something like this can be added to the module: `labels = "rata=$${COREOS_PACKET_IPV4_PUBLIC_0}"`
> 
> It is a nice solution. I see only one problem is duplication. We use 6-10 modules. And this line would be duplication.

I doubt about the nice part :-D
Terraform locals variables may help to simplify this duplication (although still duplicated, easier and less error prone).

> 
> I decided to use label `kubernetes.io/public-ipv4` similar to standard label: `kubernetes.io/hostname` and flannel annotation `flannel.alpha.coreos.com/public-ip`. Open for any suggestions.

Yeah, I think the name is the tricky part of adding this.

Kubernetes also uses beta.kubernetes.io, or other subdomains, so not sure which one we should use or even beta.kinvolk.io if there is no well-known label name yet.

In any case, I'd like to know about your use case and understand why we are doing this. What **in particular** this enables you to do?

**Comment by miry** [comment]
> In any case, I'd like to know about your use case and understand why we are doing this. What in particular this enables you to do?

Let me duplicate my comment: mostly we need to know public IPs, to automate access to the nodes.

Another case we assign Elastic IP to node, and then we can specify `nodeSelector` to run our app like nginx on machine with same IP.

**Comment by rata** [comment]
> > In any case, I'd like to know about your use case and understand why we are doing this. What in particular this enables you to do?
> 
> Let me duplicate my comment: mostly we need to know public IPs, to automate access to the nodes.

Sure, but there are tons of ways to know the IPs, for example terraform output, etc. Why a kubernetes label? Can you please elaborate on your use case that you need this as Kubernetes?

Is not that I don't think it might be useful, but if you have a concrete use case you are using this to (not just "check IPs" but elaborate on the use case, like "Check IPs and this is better to have on k8s because I don't have access to other source to read the IP addresses because of $REASON, etc.), I'd love to know it, so we can know real world use-cases and have them into account if some change is needed in the future.

I'm thinking maybe using some alpha annotation (like `alpha.kinvolk.io/public-ipv4`?) and also checking what Packet CCM does sound the most reasonable approaches. AFAIK there isnow we well-known label used for this, and I wouldn't be so lax to use the kubernetes.io domain.

What do you think?

I'll investigate those options. Feel free to check what Packet CCM does and report back if you wat too :)

> Another case we assign Elastic IP to node, and then we can specify `nodeSelector` to run our app like nginx on machine with same IP.

Not sure I follow, this label doesn't add EIP, just the node fixed IP. If you don't have a mapping of the EIP to nodes public IPs, how would this work? Can you pelase elaborate? Also, not sure you want to use pod affinity, or some other labels?, for that case.

**Comment by rata** [comment]
I've manually modified lokomotive-kubernetes and terraform-render-bootkube to use the Packet CCM (following the guide here: https://github.com/packethost/packet-ccm/tree/master/deploy/releases) and I can confirm that there are no labels added for the public IP by the CCM.

The relevant output from `kubectl describe node` when running with Packet CCM is:

```
Roles:              node
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=baremetal_0
                    beta.kubernetes.io/os=linux
                    failure-domain.beta.kubernetes.io/region=8e6470b3-b75e-47d1-bb93-45b225750975
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=rata-test-worker-worker-0
                    kubernetes.io/os=linux
                    node-role.kubernetes.io/node=
Annotations:        alpha.kubernetes.io/provided-node-ip: 10.80.134.135
                    node.alpha.kubernetes.io/ttl: 0
                    projectcalico.org/IPv4Address: 10.80.134.135/31
                    projectcalico.org/IPv4IPIPTunnelAddr: 10.2.25.1
                    volumes.kubernetes.io/controller-managed-attach-detach: true
```
There are some other labels added, though, but none for the IP.

Therefore, @miry, I suggest the following:

 * Please elaborate on the use case as much as you can (see previous comments)
 * Rename the node label to: `lokomotive.alpha.kinvolk.io/public-ipv4` (this keeps the format coresos used in annotations like `flannel.alpha.coreos.com/public-ip`)

And we are ready to merge. Also, I can do the second one if you prefer :)

miry:
@rata I have updated, pls review again

rata:
I added this as as a comment, according to what we chat:

The are several use cases:
 * In some special deployment scripts, based on kubectl, we need to ssh to the instance or grab the node public IP to record it in some other place (whitelisting access). As only kubectl is used on those scripts, having this label facilitates the workflow a lot by not having to add more things (like terraform, etc.)

 * To easily debug a node, now we can just do `kubectl describe` and then ssh to the node

@miry thanks a lot for the patch!

PR #24, opened 2019-06-20
azure: use our version of terraform-render-bootkube

invidian:

### DIFF
--- a/azure/flatcar-linux/kubernetes/bootkube.tf
+++ b/azure/flatcar-linux/kubernetes/bootkube.tf
@@ -1,6 +1,6 @@
 # Self-hosted Kubernetes assets (kubeconfig, manifests)
 module "bootkube" {
-  source = "git::https://github.com/poseidon/terraform-render-bootkube.git?ref=b96d641f6d42cf5d9bf3ac36f557aa21cc157680"
+  source = "git::https://github.com/kinvolk/terraform-render-bootkube.git?ref=d07243a9e7f6084cfe08b708731a79c26146badb"
 
   cluster_name          = "${var.cluster_name}"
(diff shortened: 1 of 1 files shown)


azure/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -1,6 +1,6 @@
 # Self-hosted Kubernetes assets (kubeconfig, manifests)
 module "bootkube" {
-  source = "git::https://github.com/poseidon/terraform-render-bootkube.git?ref=b96d641f6d42cf5d9bf3ac36f557aa21cc157680"
+  source = "git::https://github.com/kinvolk/terraform-render-bootkube.git?ref=d07243a9e7f6084cfe08b708731a79c26146badb"
```
kosyfrances, line 3: This is an older commit https://github.com/kinvolk/terraform-render-bootkube/commit/d07243a9e7f6084cfe08b708731a79c26146badb. Newer commit hash is `365b801d692804a5bbd9bb96ffff3531acb3c8d1` https://github.com/kinvolk/terraform-render-bootkube/commit/365b801d692804a5bbd9bb96ffff3531acb3c8d1

rata (approved):
LGTM.

It is not the latest commit in our repo, but it is what we are using in Packet that is well tested. So, fine for me.

We should consider to update, though, in all providers :)

invidian:
Yeah, I didn't want to run this with latest commit, but align it to other providers we support first. Update should be done separately and tested.

kosyfrances (approved):

kosyfrances:
That's a valid reason. Approved the PR.

invidian:
Tested that deployment works, so I'm merging.

kosyfrances:
And the award of the fastest PR to be shipped goes to :fire:

PR #26, opened 2019-06-20
Fix required AWS provider version

### LINKED ISSUE #25: Packet: Broken AWS provider due to version mismatch
When deploying a Packet cluster by following the [docs](https://github.com/kinvolk/lokomotive-kubernetes/blame/fc16b6175749328bf12552e1c1595b988d554326/docs/flatcar-linux/packet.md#L1), `terraform init` is currently failing with the following error:

```
Initializing modules...
- module.controller
  Getting source "git::https://github.com/kinvolk/lokomotive-kubernetes//packet/flatcar-linux/kubernetes?ref=master"
- module.worker-pool-1
  Getting source "git::https://github.com/kinvolk/lokomotive-kubernetes//packet/flatcar-linux/kubernetes/workers?ref=master"
- module.controller.bootkube
  Getting source "github.com/kinvolk/terraform-render-bootkube?ref=d07243a9e7f6084cfe08b708731a79c26146badb"

Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...
- Downloading plugin for provider "template" (1.0.0)...
- Downloading plugin for provider "tls" (1.2.0)...
- Downloading plugin for provider "packet" (1.7.2)...

No provider "aws" plugins meet the constraint "~> 1.57,~> 2.8.0".

The version constraint is derived from the "version" argument within the
provider "aws" block in configuration. Child modules may also apply
provider version constraints. To view the provider versions requested by each
module in the current configuration, run "terraform providers".

To proceed, the version constraints for this provider must be relaxed by
either adjusting or removing the "version" argument in the provider blocks
throughout the configuration.
```

This is caused by the following conflicting provider versions:

https://github.com/kinvolk/lokomotive-kubernetes/blame/fc16b6175749328bf12552e1c1595b988d554326/docs/flatcar-linux/packet.md#L58

https://github.com/kinvolk/lokomotive-kubernetes/blob/fc16b6175749328bf12552e1c1595b988d554326/packet/flatcar-linux/kubernetes/require.tf#L32

johananl:
Fixes https://github.com/kinvolk/lokomotive-kubernetes/issues/25.

`~> 1.57` was defined kind of arbitrarily. The new value is what's used by Typhoon prior to Terraform 0.12. This is also what is currently used by our AWS module.

### DIFF
--- a/packet/flatcar-linux/kubernetes/require.tf
+++ b/packet/flatcar-linux/kubernetes/require.tf
@@ -29,5 +29,5 @@ provider "packet" {
 }
 
 provider "aws" {
-  version = "~> 1.57"
+  version = ">= 1.13, < 3.0"
 }


rata (approved):
LGTM

invidian:
We should really pin to specific version to make sure things are immutable.

rata:
@invidian I agree and commented the same on previous PR (the first is before this went public, I think here: https://github.com/kinvolk/lokomotive-kubernetes-private/pull/9#discussion_r251584591).

There are two reasons that I remember:
 * Keeping those up to date might be much toil work. IMHO, it's worth the toil.
 * Keeping Typhoon convention: doesn't do it like this and to minimize diffs.

In my experience, specially with the AWS provider that changes a lot and AWS adds tons of features to their APIs, if not pin to a specific version you start to see diffs that should not exist because of new fields or something (when pinning to 1.x, not to minor releases, IIRC).

In fact, this might be a problem for reproducible builds in other tools we might want to build in the future around this terraform code. We might want to pay special care to this :)

PR #30, opened 2019-06-27
packet: Update terraform-render-bootkube

kosyfrances:
Follow up issue for other providers created here https://github.com/kinvolk/lokomotive-kubernetes/issues/29


invidian (approved):

rata (approved):
LGTM, also tested it just in case.

rata:
As a note, this is to incorpore this fix in terraform-render-bootkube: https://github.com/kinvolk/terraform-render-bootkube/pull/15

Doing it in Packet only to make sure all is fine before we upgrade all other providers (no need to go big bang)

PR #31, opened 2019-06-28
libvirt: Reference current bootkube module

pothos:
The referenced bootkube [PR](https://github.com/kinvolk/terraform-render-bootkube/pull/12) commit got merged,
so that the reference is not needed anymore
but rather would cause missing features and fixes
in the future.


invidian (commented):
We should always pin to specific version, so the outcome is fully reproducible, so I don't think this PR is needed.

PR #37, opened 2019-07-16
libvirt: Reference current bootkube module

pothos:
This updates the bootkube reference to the
current master state.
The previously referenced bootkube PR commit
got merged, so the comment is not needed anymore.

Supersedes: https://github.com/kinvolk/lokomotive-kubernetes/pull/31


invidian (approved):
LGTM

PR #45, opened 2019-07-29
Upgrade to Terraform 0.12

johananl:
## General

This is a huge PR. It is not ready to be merged yet but I'm opening it in draft mode so that people can start reviewing.

## Testing

To test the PR, install Terraform 0.12 on your machine and create a cluster using the [docs](https://github.com/kinvolk/lokomotive-kubernetes#modules).

The following platforms have been already migrated to Terraform 0.12 and can be tested:

- [ ] AWS
- [ ] Azure
- [ ] Bare metal
- [ ] Packet

## TODO

The following tasks must be completed before merging this PR:

- [ ] Migrate DO, GCP and KVM platforms to Terraform 0.12.
- [ ] Fix whitespace and indentation issues created by `terraform 0.12upgrade`.
- [ ] Finish updating docs.
- [ ] Clean up version requirements.
- [ ] Update CI process to support Terraform 0.12.
- [ ] Merge https://github.com/kinvolk/terraform-render-bootkube/tree/tf12.
- [ ] Pin terraform-render-bootkube module reference to the correct commit.
- [ ] Rebase and squash commits nicely.

The following follow up tasks should be opened after merging:

- [ ] Update lokoctl to support Terraform 0.12.

### DIFF
--- a/aws/flatcar-linux/kubernetes/require.tf
+++ b/aws/flatcar-linux/kubernetes/require.tf
@@ -23,3 +23,4 @@ provider "template" {
 provider "tls" {
   version = "~> 2.0.1"
 }
+
--- a/aws/flatcar-linux/kubernetes/require.tf
+++ b/aws/flatcar-linux/kubernetes/require.tf
@@ -17,7 +17,7 @@ provider "null" {
 }
 
 provider "template" {
-  version = "~> 1.0"
+  version = "~> 2.1.2"
 }
 
 provider "tls" {
--- a/aws/flatcar-linux/kubernetes/require.tf
+++ b/aws/flatcar-linux/kubernetes/require.tf
@@ -21,5 +21,5 @@ provider "template" {
 }
 
 provider "tls" {
-  version = "~> 1.0"
+  version = "~> 2.0.1"
 }
--- a/aws/flatcar-linux/kubernetes/require.tf
+++ b/aws/flatcar-linux/kubernetes/require.tf
@@ -13,7 +13,7 @@ provider "local" {
 }
 
 provider "null" {
-  version = "~> 1.0"
+  version = "~> 2.1.2"
 }
 
 provider "template" {
--- a/bare-metal/flatcar-linux/kubernetes/groups.tf
+++ b/bare-metal/flatcar-linux/kubernetes/groups.tf
@@ -1,33 +1,51 @@
 resource "matchbox_group" "install" {
-  count = "${length(var.controller_names) + length(var.worker_names)}"
+  count = length(var.controller_names) + length(var.worker_names)
 
-  name = "${format("install-%s", element(concat(var.controller_names, var.worker_names), count.index))}"
+  name = format(
+    "install-%s",
+    element(concat(var.controller_names, var.worker_names), count.index),
+  )
 
-  profile = "${local.flavor == "flatcar" ? var.cached_install == "true" ? element(matchbox_profile.cached-flatcar-linux-install.*.name, count.index) : element(matchbox_profile.flatcar-install.*.name, count.index) : var.cached_install == "true" ? element(matchbox_profile.cached-container-linux-install.*.name, count.index) : element(matchbox_profile.container-linux-install.*.name, count.index)}"
+  profile = local.flavor == "flatcar" ? var.cached_install == "true" ? element(
+    matchbox_profile.cached-flatcar-linux-install.*.name,
+    count.index,
+    ) : element(matchbox_profile.flatcar-install.*.name, count.index) : var.cached_install == "true" ? element(
+    matchbox_profile.cached-container-linux-install.*.name,
+    count.index,
+  ) : element(matchbox_profile.container-linux-install.*.name, count.index)
 
   selector = {
-    mac = "${element(concat(var.controller_macs, var.worker_macs), count.index)}"
+    mac = element(concat(var.controller_macs, var.worker_macs), count.index)
   }
 }
 
 resource "matchbox_group" "controller" {
-  count   = "${length(var.controller_names)}"
-  name    = "${format("%s-%s", var.cluster_name, element(var.controller_names, count.index))}"
-  profile = "${element(matchbox_profile.controllers.*.name, count.index)}"
+  count = length(var.controller_names)
+  name = format(
+    "%s-%s",
+    var.cluster_name,
+    element(var.controller_names, count.index),
+  )
+  profile = element(matchbox_profile.controllers.*.name, count.index)
 
   selector = {
-    mac = "${element(var.controller_macs, count.index)}"
+    mac = element(var.controller_macs, count.index)
     os  = "installed"
   }
 }
 
 resource "matchbox_group" "worker" {
-  count   = "${length(var.worker_names)}"
-  name    = "${format("%s-%s", var.cluster_name, element(var.worker_names, count.index))}"
-  profile = "${element(matchbox_profile.workers.*.name, count.index)}"
+  count = length(var.worker_names)
+  name = format(
+    "%s-%s",
+    var.cluster_name,
+    element(var.worker_names, count.index),
+  )
+  profile = element(matchbox_profile.workers.*.name, count.index)
 
   selector = {
-    mac = "${element(var.worker_macs, count.index)}"
+    mac = element(var.worker_macs, count.index)
     os  = "installed"
   }
 }
+
--- a/bare-metal/flatcar-linux/kubernetes/groups.tf
+++ b/bare-metal/flatcar-linux/kubernetes/groups.tf
@@ -3,19 +3,13 @@ resource "matchbox_group" "install" {
 
   name = format(
     "install-%s",
-    element(concat(var.controller_names, var.worker_names), count.index),
+    concat(var.controller_names, var.worker_names)[count.index]
   )
 
-  profile = local.flavor == "flatcar" ? var.cached_install == "true" ? element(
-    matchbox_profile.cached-flatcar-linux-install.*.name,
-    count.index,
-    ) : element(matchbox_profile.flatcar-install.*.name, count.index) : var.cached_install == "true" ? element(
-    matchbox_profile.cached-container-linux-install.*.name,
-    count.index,
-  ) : element(matchbox_profile.container-linux-install.*.name, count.index)
+  profile = local.flavor == "flatcar" ? var.cached_install == "true" ? matchbox_profile.cached-flatcar-linux-install[count.index].name : matchbox_profile.flatcar-install[count.index].name : var.cached_install == "true" ? matchbox_profile.cached-container-linux-install[count.index].name : matchbox_profile.container-linux-install[count.index].name
 
   selector = {
-    mac = element(concat(var.controller_macs, var.worker_macs), count.index)
+    mac = concat(var.controller_macs, var.worker_macs)[count.index]
   }
 }
 
@@ -24,12 +18,12 @@ resource "matchbox_group" "controller" {
   name = format(
     "%s-%s",
     var.cluster_name,
-    element(var.controller_names, count.index),
+    var.controller_names[count.index]
   )
-  profile = element(matchbox_profile.controllers.*.name, count.index)
+  profile = matchbox_profile.controllers[count.index].name
 
   selector = {
-    mac = element(var.controller_macs, count.index)
+    mac = var.controller_macs[count.index]
     os  = "installed"
   }
 }
@@ -39,12 +33,12 @@ resource "matchbox_group" "worker" {
   name = format(
     "%s-%s",
     var.cluster_name,
-    element(var.worker_names, count.index),
+    var.worker_names[count.index]
   )
-  profile = element(matchbox_profile.workers.*.name, count.index)
+  profile = matchbox_profile.workers[count.index].name
 
   selector = {
-    mac = element(var.worker_macs, count.index)
+    mac = var.worker_macs[count.index]
     os  = "installed"
   }
 }
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -87,7 +87,7 @@ provider "tls" {
 }
 
 provider "packet" {
-  version = "~> 1.2"
+  version = "~> 2.2.1"
   alias   = "default"
 }
 ```
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -82,7 +82,7 @@ provider "template" {
 }
 
 provider "tls" {
-  version = "~> 1.0"
+  version = "~> 2.0.1"
   alias   = "default"
 }
 
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -72,7 +72,7 @@ provider "local" {
 }
 
 provider "null" {
-  version = "~> 1.0"
+  version = "~> 2.1.2"
   alias   = "default"
 }
 
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -176,7 +176,7 @@ module "worker-pool-helium" {
   facility     = "ams1"
   pool_name    = "helium"
 
-  count = 2
+  worker_count = 2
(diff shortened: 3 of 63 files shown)


packet/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -16,7 +16,7 @@ variable "pool_name" {
   description = "Unique worker pool name (prepended to hostname)"
 }
 
-variable "count" {
+variable "worker_count" {
```
invidian, line 19: This also needs to be updated in examples in documentation.

packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -1,105 +1,112 @@
...
+  facilities       = [var.facility]
   operating_system = "custom_ipxe"
   billing_cycle    = "hourly"
-  project_id       = "${var.project_id}"
-  ipxe_script_url  = "${var.ipxe_script_url}"
+  project_id       = var.project_id
+  ipxe_script_url  = var.ipxe_script_url
   always_pxe       = "false"
-  user_data        = "${element(data.ct_config.controller-install-ignitions.*.rendered, count.index)}"
+  user_data = element(
```
invidian: I think this can be simplified to `data.ct_config.controller-install-ignitions[count.index].rendered`.

packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -1,105 +1,112 @@
...
+  zone_id = var.dns_zone_id
 
-  name = "${format("%s-etcd%d.%s.", var.cluster_name, count.index, var.dns_zone)}"
+  name = format("%s-etcd%d.%s.", var.cluster_name, count.index, var.dns_zone)
   type = "A"
   ttl  = 300
 
   # private IPv4 address for etcd
-  records = ["${element(packet_device.controllers.*.access_private_ipv4, count.index)}"]
+  records = [element(packet_device.controllers.*.access_private_ipv4, count.index)]
```
invidian: `packet_device.controllers[count.index].access_private_ipv4`?

packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -1,105 +1,112 @@
...
+    data.ct_config.controller-install-ignitions.*.rendered,
+    count.index,
+  )
 }
 
 data "ct_config" "controller-install-ignitions" {
-  count   = "${var.controller_count}"
-  content = "${element(data.template_file.controller-install.*.rendered, count.index)}"
+  count = var.controller_count
+  content = element(
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -1,105 +1,112 @@
...
-    os_channel           = "${var.os_channel}"
-    os_version           = "${var.os_version}"
+  vars = {
+    os_channel           = var.os_channel
+    os_version           = var.os_version
     flatcar_linux_oem    = "packet"
-    ssh_keys             = "${jsonencode("${var.ssh_keys}")}"
-    postinstall_ignition = "${element(data.ct_config.controller-ignitions.*.rendered, count.index)}"
+    ssh_keys             = jsonencode(var.ssh_keys)
+    postinstall_ignition = element(data.ct_config.controller-ignitions.*.rendered, count.index)
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -1,105 +1,112 @@
...
   }
 }
 
 data "ct_config" "controller-ignitions" {
-  count    = "${var.controller_count}"
+  count    = var.controller_count
   platform = "packet"
-  content  = "${element(data.template_file.controller-configs.*.rendered, count.index)}"
+  content = element(
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -1,46 +1,46 @@
 # Secure copy etcd TLS assets to controllers.
 resource "null_resource" "copy-controller-secrets" {
-  count = "${var.controller_count}"
+  count = var.controller_count
 
   connection {
     type    = "ssh"
-    host    = "${element(packet_device.controllers.*.access_public_ipv4, count.index)}"
+    host    = element(packet_device.controllers.*.access_public_ipv4, count.index)
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -95,31 +95,35 @@ resource "null_resource" "bootkube-start" {
...
 data "template_file" "controller_host_endpoints" {
-  count    = "${var.controller_count}"
-  template = "${file("${path.module}/calico/controller-host-endpoint.yaml.tmpl")}"
+  count    = var.controller_count
+  template = file("${path.module}/calico/controller-host-endpoint.yaml.tmpl")
 
-  vars {
-    node_name = "${element(packet_device.controllers.*.hostname, count.index)}"
+  vars = {
+    node_name = element(packet_device.controllers.*.hostname, count.index)
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -95,31 +95,35 @@ resource "null_resource" "bootkube-start" {
...
 data "template_file" "worker_host_endpoints" {
-  count    = "${var.worker_count}"
-  template = "${file("${path.module}/calico/worker-host-endpoint.yaml.tmpl")}"
+  count    = var.worker_count
+  template = file("${path.module}/calico/worker-host-endpoint.yaml.tmpl")
 
-  vars {
-    node_name = "${element("${var.worker_nodes_hostnames}", count.index)}"
+  vars = {
+    node_name = element(var.worker_nodes_hostnames, count.index)
```
invidian: Ditto array notation.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -95,31 +95,35 @@ resource "null_resource" "bootkube-start" {
...
-  template = "${file("${path.module}/calico/host-protection.yaml.tmpl")}"
+  template = file("${path.module}/calico/host-protection.yaml.tmpl")
 
   vars = {
-    controller_host_endpoints = "${join("\n", data.template_file.controller_host_endpoints.*.rendered)}"
-    worker_host_endpoints     = "${join("\n", data.template_file.worker_host_endpoints.*.rendered)}"
-    management_cidrs          = "${jsonencode("${var.management_cidrs}")}"
-    cluster_internal_cidrs    = "${jsonencode(list("${var.node_private_cidr}", "${var.pod_cidr}", "${var.service_cidr}"))}"
-    etcd_server_cidrs         = "${jsonencode("${packet_device.controllers.*.access_private_ipv4}")}"
+    controller_host_endpoints = join(
```
invidian: Can we revert formatting change here?

packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -1,57 +1,62 @@
...
-  ipxe_script_url  = "${var.ipxe_script_url}"
+  project_id       = var.project_id
+  ipxe_script_url  = var.ipxe_script_url
   always_pxe       = "false"
-  user_data        = "${data.ct_config.install-ignitions.rendered}"
+  user_data        = data.ct_config.install-ignitions.rendered
 
   # If not present in the map, it uses ${var.reservation_ids_default}
-  hardware_reservation_id = "${lookup(var.reservation_ids, format("worker-%v", count.index), var.reservation_ids_default)}"
+  hardware_reservation_id = lookup(
```
invidian: Ditto formatting.

invidian (changes requested):
The suggested changes for Packet provider should also be applied for other providers.

I didn't test it yet.

Also, we need to update the CI as well to support new versions of providers and Terraform.

mauriciovasquezbernal:
Could we use terraform 0.12 types (https://www.terraform.io/docs/configuration/types.html) to avoid problems like https://github.com/kinvolk/lokomotive-kubernetes/pull/44?

johananl:
>Could we use terraform 0.12 types (https://www.terraform.io/docs/configuration/types.html) to avoid problems like #44?

What change are you suggesting @mauriciovasquezbernal? Could you post an example or even a diff?

packet/flatcar-linux/kubernetes/workers/variables.tf
johananl: Oops, done.

packet/flatcar-linux/kubernetes/controllers.tf
johananl: ~That syntax is invalid.~

Had a typo. Fixed.

invidian:
Will review tomorrow. I was thinking, is there a way to configure Terraform version used for running tests here instead of in CI code? I think this would be good direction to move in the future, so we can make things more self contained.

docs/flatcar-linux/packet.md
```diff
@@ -87,7 +87,7 @@ provider "tls" {
 }
 
 provider "packet" {
-  version = "~> 1.2"
+  version = "~> 2.2.1"
```
invidian, line 90: `2.3.0` is available now.

aws/flatcar-linux/kubernetes/require.tf
```diff
@@ -21,5 +21,5 @@ provider "template" {
 }
 
 provider "tls" {
-  version = "~> 1.0"
+  version = "~> 2.0.1"
```
invidian, line 24: `2.1.0` is now available.

packet/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -1,5 +1,5 @@
 module "bootkube" {
-  source = "github.com/kinvolk/terraform-render-bootkube?ref=0913331140747fd464946981c1d43a895d7a630b"
+  source = "github.com/kinvolk/terraform-render-bootkube?ref=9c1ff5a9e8b3e98922980a097c49c8b3a903437a"
```
invidian, line 2: I think updates should cascade, we should merge bootkube PR first, then this one.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -99,31 +99,35 @@ resource "null_resource" "bootkube-start" {
...
+      "\n",
+      data.template_file.controller_host_endpoints.*.rendered,
+    )
+    worker_host_endpoints  = join("\n", data.template_file.worker_host_endpoints.*.rendered)
+    management_cidrs       = jsonencode(var.management_cidrs)
+    cluster_internal_cidrs = jsonencode([var.node_private_cidr, var.pod_cidr, var.service_cidr])
+    etcd_server_cidrs      = jsonencode(packet_device.controllers.*.access_private_ipv4)
   }
 }
+
```
invidian, line 130: I would get rid of those empty lines at the end of every file.

bare-metal/flatcar-linux/kubernetes/groups.tf
```diff
@@ -1,33 +1,51 @@
 resource "matchbox_group" "install" {
-  count = "${length(var.controller_names) + length(var.worker_names)}"
+  count = length(var.controller_names) + length(var.worker_names)
 
-  name = "${format("install-%s", element(concat(var.controller_names, var.worker_names), count.index))}"
+  name = format(
```
invidian, line 4: Formatting changed could/should be reverted here.

bare-metal/flatcar-linux/kubernetes/groups.tf
```diff
@@ -1,33 +1,51 @@
...
+    "install-%s",
+    element(concat(var.controller_names, var.worker_names), count.index),
+  )
 
-  profile = "${local.flavor == "flatcar" ? var.cached_install == "true" ? element(matchbox_profile.cached-flatcar-linux-install.*.name, count.index) : element(matchbox_profile.flatcar-install.*.name, count.index) : var.cached_install == "true" ? element(matchbox_profile.cached-container-linux-install.*.name, count.index) : element(matchbox_profile.container-linux-install.*.name, count.index)}"
+  profile = local.flavor == "flatcar" ? var.cached_install == "true" ? element(
```
invidian: Ditto formatting + please not so many tenary operators! :smile:

invidian (changes requested):
I commented on some nits I would fix, especially formatting in bare-metal module could be improved.

I also think we should merge the PR into bootkube first, then point to the commit from master branch here.

Overall LGMT, great work @johananl.

invidian:
> I was thinking, is there a way to configure Terraform version used for running tests here instead of in CI code? I think this would be good direction to move in the future, so we can make things more self contained.

Created internal ticket for that.

johananl:
>I commented on some nits I would fix, especially formatting in bare-metal module could be improved.

@invidian the focus of this PR is **switching to Terraform 0.12**. I think we should stay focused on that and not slip to an overall refactor. I am well aware of the things you've pointed out inline and have spotted most of them myself, I'm just trying to stay focused on switching to the newer Terraform version. For example, I've addressed your previous comment regarding using array notation instead of `element()` - and it took me 2 hours to fix all of these throughout the codebase and test that I didn't break anything. Furthermore, this is a remnant from Typhoon and not something the migration introduced. That said, I think we can fix a few small things but I'd rather not turn this into a full-blown refactor (something we can definitely do in a separate PR *after* the migration).

>I also think we should merge the PR into bootkube first, then point to the commit from master branch here.

It's written as a TODO item in the description.

invidian:
>@invidian the focus of this PR is switching to Terraform 0.12. I think we should stay focused on that and not slip to an overall refactor. I am well aware of the things you've pointed out inline and have spotted most of them myself, I'm just trying to stay focused on switching to the newer Terraform version. For example, I've addressed your previous comment regarding using array notation instead of element() - and it took me 2 hours to fix all of these throughout the codebase and test that I didn't break anything. Furthermore, this is a remnant from Typhoon and not something the migration introduced. That said, I think we can fix a few small things but I'd rather not turn this into a full-blown refactor (something we can definitely do in a separate PR after the migration).

I understand and accept that some thing might be out of scope for this PR.

I think `element()` thingy should be considered as an improvement, which migration allows us to do, so I would treat it as optional, but it's good that you've made it.

 On the other hand styling is directly caused by the migration (broken tool on Terraform side), so I consider making formatting more inconsistent as an regression here, which shouldn't be introduced. I don't leaving such things as TODO later, because usually it's hard to go back to such things.

>I'd rather not turn this into a full-blown refactor (something we can definitely do in a separate PR after the migration).

If you refer to: `please not so many tenary operators! smile`, I'm sorry for the confusion, I just spotted chained tenary operators and found it funny, I didn't want to suggest, that this should be fixed.

Bottom line is, if you think the patch is working and we should merge it,so we can move forward, and should be improved later, I don't insist on any of my comments.

mauriciovasquezbernal:
> > Could we use terraform 0.12 types (https://www.terraform.io/docs/configuration/types.html) to avoid problems like #44?
> 
> What change are you suggesting @mauriciovasquezbernal? Could you post an example or even a diff?

AFAIK terraform 0.12 includes a new support for type constraints, we can mark some variables as `number` or `bool`. For instance, `setup_raid` should have `type = bool`, this could avoid problems like  in https://github.com/kinvolk/lokomotive-kubernetes/pull/44 

https://github.com/kinvolk/lokomotive-kubernetes/blob/d454078bd2a61c0624dae65efbf4233c8e17ef7e/packet/flatcar-linux/kubernetes/workers/variables.tf#L102-L106

pothos:
Please rebase to pick up the CI configuration files.

alvelcom:
Supporting Terraform v0.12 is a very important thing 👍. Is it still on the table?

rata:
Yes, of course! But we've been quite slow moving forward with this lately :(

invidian:
I'll close this one. Another attempt in #119.

invidian:
@johananl please remove the branch if you think it's fine.

PR #46, opened 2019-07-31
 packet: first stab at cluster autoscaler 

iaguis:
This adds a `cluster_autoscaler` option to the packet terraform module.

When set to true, it will add and configure the Packet Cluster
Autoscaler in the resulting cluster, which extends or shrinks the worker
node pool depending on the Pod load.

This also adds a number of new variables, some should be taken from the
worker module so they are added as outputs there.

* `worker_user_data`: it's needed for the autoscaler to spawn new nodes,
should be taken from the worker pool module.
* `àuth_token`: the autoscaler needs the auth token to spawn new nodes.
* `worker_type`, `worker_os_channel`: to configure the nodes that will
be spawned by the autoscaler.
* `pool_name`: autoscaler pool name, for now only one pool is supported
and should be taken from the worker module.
* `min_workers`, `mas_workers`: limits for the total number of workers
when using the autoscaler.

Worker nodes created when the cluster is spawned are also part of
the autoscaler pool and are tagged accordingly.

I decided to consider it a core component and add it at cluster set up
but this is up for discussion.

The autoscaler container image is not the `mist` one because of one bug
I found when scaling up the cluster by more than one node. I commented
on the upstrem PR so hopefully this gets fixed in their image soon.

Note that HostEndpoint objects that secure worker nodes are not applied
here, this needs to be solved later.

This is based on #28


packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -94,6 +101,68 @@ resource "null_resource" "bootkube-start" {
   }
 }
 
+resource "null_resource" "autoscaler-config" {
+  count = "${var.enable_autoscaler == "true" ? 1 : 0}"
+
+  depends_on = [
```
invidian, line 107: I think `depends_on` should go at the bottom of the resource.

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -127,3 +127,52 @@ variable "enable_aggregation" {
...
+  type        = "string"
+}
+
+variable "worker_os_channel" {
+  description = "Flatcar Linux channel to install from (stable, beta, alpha, edge) on workers"
+  default     = "stable"
+  type        = "string"
+}
+
+variable "enable_autoscaler" {
```
invidian, line 155: Shouldn't that be named `worker_enable_autoscaler`? What if we want to have autoscaling enabled for controlplane later on?

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -127,3 +127,52 @@ variable "enable_aggregation" {
...
+  description = "Worker pool name"
+  default     = ""
+  type        = "string"
+}
+
+
+variable "min_workers" {
+  description = "Minimum number of workers for the autoscaler"
+  default     = "1"
+  type        = "string"
```
invidian, line 171: Why string?

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -127,3 +127,52 @@ variable "enable_aggregation" {
...
+variable "worker_os_channel" {
+  description = "Flatcar Linux channel to install from (stable, beta, alpha, edge) on workers"
+  default     = "stable"
+  type        = "string"
+}
+
+variable "enable_autoscaler" {
+  description = "Enable cluster autoscaler (defaults to false)"
+  default     = "false"
+  type        = "string"
```
invidian, line 158: Why string?

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -127,3 +127,52 @@ variable "enable_aggregation" {
...
+}
+
+variable "pool_name" {
+  description = "Worker pool name"
+  default     = ""
```
invidian, line 163: Can this line be removed? Or does it make the variable required?

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -94,6 +101,68 @@ resource "null_resource" "bootkube-start" {
...
+
+data "template_file" "autoscaler_secret" {
+  template = "${file("${path.module}/autoscaler/cluster-autoscaler-secret.yaml.tmpl")}"
+
+  vars {
+    facility = "${var.facility}"
+    os_channel = "${var.worker_os_channel}"
+    type = "${var.worker_type}"
+    user_data = "${base64encode(var.worker_user_data)}"
+    packet_auth_token = "${base64encode(var.auth_token)}"
```
invidian, line 161: We actually wanted to remove `auth_token` variable, to rely on Terraform provider authentication method.

packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -35,3 +38,4 @@ data "template_file" "configs" {
     taints                = "${var.taints}"
   }
 }
+
```
invidian, line 41: Extra line not needed.

packet/flatcar-linux/kubernetes/autoscaler/cluster-autoscaler-deployment.yaml.tmpl
```diff
@@ -0,0 +1,166 @@
...
+                  name: cluster-autoscaler-packet
+                  key: authtoken
+          command:
+            - ./cluster-autoscaler
+            - --alsologtostderr
+            - --cluster-name=${cluster_name}
+            - --cloud-config=/config/cloud-config
+            - --cloud-provider=packet
+            - --nodes=${min_workers}:${max_workers}:${pool_name}
+            - --scale-down-unneeded-time=1m0s
```
invidian, line 155: What is the billing timeframe on Packet? Is it hourly or per minute? Because if it's per hour, on every node spawn you will be charged for the entire hour. I think also provisioning time should be considered here, to avoid nodes being removed before they get a chance to appear in the cluster.

invidian (changes requested):
Commented only on last commit in the PR, since this seems to be autoscaling part.

The Terraform changes looks good, however, I think that autoscaling manifests should be moved to component, because:
- cluster deployment should maintain minimal set of deployments to make cluster running
- we want to support autoscaling on different platforms as well, which in this case make duplicate code, which will be harder to maintain

iaguis:
We've decided on a non-terraform approach. Closing.

PR #59, opened 2019-09-02
Add CI cluster configurations for AWS and Packet

pothos:
Previously the configuration was part of the CI
pipelines which caused problems when code changed
in a pull request.

By including the CI configurations in the repository
both code and CI configuration can be changed at once.

### DIFF
--- a/ci/aws/aws-cluster.tf.envsubst
+++ b/ci/aws/aws-cluster.tf.envsubst
@@ -1,0 +1,21 @@
+module "aws-tempest" {
+  source = "../../aws/flatcar-linux/kubernetes"
+
+  providers = {
+    aws = "aws.default"
+    local = "local.default"
+    null = "null.default"
+    template = "template.default"
+    tls = "tls.default"
+  }
+
+  cluster_name = "$CLUSTER_ID"
+  dns_zone = "$AWS_DNS_ZONE"
+  dns_zone_id = "$AWS_DNS_ZONE_ID"
+  ssh_authorized_key = "$PUB_KEY"
+
+  asset_dir = "${pathexpand("~/lokomotive-assets")}"
+
+  worker_count = 2
+  worker_type  = "t3.small"
+}
--- a/ci/packet/packet-cluster.tf.envsubst
+++ b/ci/packet/packet-cluster.tf.envsubst
@@ -1,0 +1,62 @@
+module "controller" {
+  source = "../../packet/flatcar-linux/kubernetes"
+
+  providers = {
+    aws      = "aws.default"
+    local    = "local.default"
+    null     = "null.default"
+    template = "template.default"
+    tls      = "tls.default"
+    packet   = "packet.default"
+  }
+
+  dns_zone = "$AWS_DNS_ZONE"
+  dns_zone_id = "$AWS_DNS_ZONE_ID"
+
+  ssh_keys = ["$PUB_KEY"]
+
+  asset_dir = "${pathexpand("~/assets")}"
+  cluster_name = "$CLUSTER_ID"
+  project_id = "$PACKET_PROJECT_ID"
+
+  facility = "sjc1"
+
+  worker_count = 2
+  worker_nodes_hostnames = "${concat("${module.worker-pool-1.worker_nodes_hostname}")}"
+
+  controller_count = 1
+  controller_type  = "t1.small.x86"
+
+  management_cidrs = [
+    "0.0.0.0/0",       # Instances can be SSH-ed into from anywhere on the internet.
+  ]
+
+  node_private_cidr = "10.0.0.0/8"
+}
+
+module "worker-pool-1" {
+  source = "../../packet/flatcar-linux/kubernetes/workers"
+
+  providers = {
+    local    = "local.default"
+    template = "template.default"
+    tls      = "tls.default"
+    packet   = "packet.default"
+  }
+
+
+  ssh_keys = ["$PUB_KEY"]
+
+  cluster_name = "$CLUSTER_ID"
+  project_id = "$PACKET_PROJECT_ID"
+
+  facility     = "sjc1"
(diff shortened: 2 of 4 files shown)


ci/packet/packet-cluster.tf.envsubst
```diff
@@ -0,0 +1,62 @@
...
+  ssh_keys = ["$PUB_KEY"]
+
+  asset_dir = "${pathexpand("~/assets")}"
+  cluster_name = "$CLUSTER_ID"
+  project_id = "$PACKET_PROJECT_ID"
+
+  facility = "sjc1"
+
+  worker_count = 2
+  worker_nodes_hostnames = "${concat("${module.worker-pool-1.worker_nodes_hostname}")}"
```
rata, line 25: No need for concat, there is only one list there :)

ci/packet/packet-cluster.tf.envsubst
```diff
@@ -0,0 +1,62 @@
...
+
+  controller_count = 1
+  controller_type  = "t1.small.x86"
+
+  management_cidrs = [
+    "0.0.0.0/0",       # Instances can be SSH-ed into from anywhere on the internet.
+  ]
+
+  node_private_cidr = "10.0.0.0/8"
```
rata, line 34: Comment explaining here that, as of writing, the node private cidr is always included in this range and facilitates testing? Maybe also elaborate on why we can't specify the more specific cidr (because is not defined until at least one server is created and if all servers are deleted after that it is still not fixed)

ci/packet/packet-cluster.tf.envsubst
```diff
@@ -0,0 +1,62 @@
...
+
+  facility     = "sjc1"
+  pool_name    = "pool-1"
+
+  count = 2
+  type  = "t1.small.x86"
+
+  kubeconfig = "${module.controller.kubeconfig}"
+
+  labels = "node.supernova.io/role=backend,node-role.kubernetes.io/backend="
```
rata, line 61: ? Is this to test labels? Or why are we adding this?

ci/packet/packet-cluster.tf.envsubst
```diff
@@ -0,0 +1,62 @@
...
+    tls      = "tls.default"
+    packet   = "packet.default"
+  }
+
+  dns_zone = "$AWS_DNS_ZONE"
+  dns_zone_id = "$AWS_DNS_ZONE_ID"
+
+  ssh_keys = ["$PUB_KEY"]
+
+  asset_dir = "${pathexpand("~/assets")}"
```
rata, line 18: Why use different asset directories in aws in packet? If it's for overlap, doesn't it make sense to include the provider in tha path instead?

ci/packet/packet-cluster.tf.envsubst
```diff
@@ -0,0 +1,62 @@
...
+  cluster_name = "$CLUSTER_ID"
+  project_id = "$PACKET_PROJECT_ID"
+
+  facility = "sjc1"
```
rata, line 22: We can use locals to avoid duplicating this here and in the worker module (or env vars, as we prefer)

ci/aws/aws-cluster.tf.envsubst
```diff
@@ -0,0 +1,21 @@
...
+
+  cluster_name = "$CLUSTER_ID"
+  dns_zone = "$AWS_DNS_ZONE"
+  dns_zone_id = "$AWS_DNS_ZONE_ID"
+  ssh_authorized_key = "$PUB_KEY"
+
+  asset_dir = "${pathexpand("~/lokomotive-assets")}"
+
+  worker_count = 2
+  worker_type  = "t3.small"
```
rata, line 20: In Packet we are defining the controllers (number and size) and that matches the default. I thought we wanted to be explicit, just in case the default changes or something.

Should we add the controllers size and count here, too? Or remove it on Packet?

IMHO, it makes sense to be explicit and specify them, so tests don't depend on the value of some defaults and we test exactly what we want to test

rata (commented):
Some small comments, but mostly LGTM from the coding POV.

Some questions:
 1. The idea, then, to keep this file up to date with recent changes, so is also easy to update the CI if needed?

2. This might be ok for now, but this basically means that we are breaking existing users (if old code can't be used with newer lokomotive kubernetes versions). That is, we are not being backwards compatible. Or am I missing something?

I think we should create an issue to discuss this in the future (out of scope for this PR). We should consider having releases and that stuff, and write something about backwards compatibility (IMHO, it should NOT be broken after a platform is stable, at least, unless when stated and is done carefully, of course).

I guess you created this because some recent change broke the CI? Can you post the changes that broke it so we can have examples of this and understand if this was really needed?

pothos:
Hey,

thanks for the comments. I just moved the config and I think it's better to change/update it in a separate PR when it the changes are actually tested by CI.

pothos:
Yes, the idea is to coordinate changing the needed CI changes easily and have them apply only to the PR that changes it (as long as it is not merged).

About the breaking change: The provider versions need to be updated in https://github.com/kinvolk/lokomotive-kubernetes/pull/58.

ci/packet/packet-cluster.tf.envsubst
pothos: No special meaning, overlap is not a problem because the pipelines run in containers.

ci/packet/packet-cluster.tf.envsubst
pothos: They were used in the documentation, so I just took them so that labels are tested.

ci/packet/packet-cluster.tf.envsubst
pothos: Yes, it's just inherited. Maybe we want to test multiple pools since we need them at least for the VPN?

PR #60, opened 2019-09-02
packet: add controller resertavions ids

rata:
Commit 9211118c75b0f3463d616bbcddea07af4c38aee1 and 
b16e99c182f9c3da1905d649d3e487607f73347a already added hardware
reservation id support for workers. This patch is very similar, but it
adds support for controllers.

The motivation behind is the same as for workers (use reserved instances
is always nice, specially for controllers that will always be running)
added to that with reservations you can also guarantee that controllers,
for example, are isolated (different racks, etc.) as much as possible in
the Packet facility for better resilency.

We've seen issues in the past were controllers landed behind the same
Packet switch and all the control plane was down due to an issue in the 
switch.


invidian (approved):
LGTM, seems reasonable.

PR #61, opened 2019-09-04
Update the libvirt platform to use fw_cfg_name

pothos:
Recently Flatcar changed to expect the ignition
config under the name org.flatcar-linux.
The terraform libvirt provider has a commit in
its master branch that allows to set this option.
Use it and document how to build from source for
the mean time.


dongsupark (approved):
Ah, https://github.com/dmacvicar/terraform-provider-libvirt/pull/627 was finally merged.
Looks good from a quick look.

PR #63, opened 2019-09-05
Document changeable Packet variables

pothos:


docs/flatcar-linux/packet.md
```diff
@@ -326,3 +326,13 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
...
 
 Documentation about Packet hardware reservation id can be found here: https://support.packet.com/kb/articles/reserved-hardware.
+
+## Post-installation modification
+
+Currently the only tested ways to modify a cluster are:
+
+* Adding new worker pools, done by adding a new worker module and concating its `.worker_nodes_hostname` to the `worker_nodes_hostnames` list of the controller module
+* Scaling a worker pool by changing the `count` to delete or add nodes, even to 0 (but the worker pool definition has to be kept)
+* Changing the instance type of a worker pool by altering `type`, e.g., from `t1.small.x86` to `c1.small.x86`, which will recreate the nodes
```
rata: I'll be more explicit that nodes won't be recreated in any order, they will be destroyed first and then created again so it may create downtime.

rata (approved):
LGTM, small comment that doesn't need to be reviewd again after is fixed :)

docs/flatcar-linux/packet.md
pothos: Good point.

docs/flatcar-linux/packet.md
```diff
@@ -326,3 +326,13 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
...
+
+## Post-installation modification
+
+Currently the only tested ways to modify a cluster are:
+
+* Adding new worker pools, done by adding a new worker module and concating its `.worker_nodes_hostname` to the `worker_nodes_hostnames` list of the controller module.
+* Scaling a worker pool by changing the `count` to delete or add nodes, even to 0 (but the worker pool definition has to be kept and the total number of workers must be > 0).
+* Changing the instance type of a worker pool by altering `type`, e.g., from `t1.small.x86` to `c1.small.x86`, which will recreate the nodes, causing downtime since they are destroyed first and then created again.
+
+This list may be expanded in the future but for now other changes are not supported but can be done on your own risk.
```
rata: nit: I thik is *at* your own risk, not *on*. But not important :)

pothos: fixed

PR #64, opened 2019-09-06
Copy kubeconfig via SSH to reduce AWS Ignition file size

pothos:
Sometimes the AWS Ignition file for the controller hit the
size limit of 16 kB, thus the cluster could not be created.

Move the big kubeconfig out of the controller Ignition file by
copying it through SSH. It contains more scripts than the worker
Ignition file. This brings the controller Ignition down to ~10 kB
while the worker stays as ~11 kB. If needed, the worker kubeconfig
can later also be moved out by copying it through SSH but for now
both Ignition files have each ~5 kB free for encoded data.

### DIFF
--- a/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
@@ -51,6 +51,7 @@ systemd:
         Description=Kubelet via Hyperkube
         Wants=rpc-statd.service
         [Service]
+        ConditionPathExists=/etc/kubernetes/kubeconfig
         EnvironmentFile=/etc/kubernetes/kubelet.env
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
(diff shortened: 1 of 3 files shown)


rata (approved):
LGTM. Do you think it makes sense to do it for other platforms too? Maybe we can create an follow-up issue for that? :)

pothos:
Other platforms don't have the size limit. If we had code sharing between all platforms, then we would need to do it for all, but since it's not the case, they can stay as they are (unless it's somehow more secure to copy a secret via SSH than using Ignition but I don't think so).

invidian (commented):
LGTM, testing it now to see if it solves my issue.

invidian (approved):
LGTM, works

rata:
> Other platforms don't have the size limit. If we had code sharing between all platforms, then we would need to do it for all, but since it's not the case, they can stay as they are (unless it's somehow more secure to copy a secret via SSH than using Ignition but I don't think so).

Oh, if the limit is not a problem on others (sorry!) is not a big deal. I think having the same code was nice, too, but we can change it later if we prefer that (I thought it was needed, sorry :-D)

PR #70, opened 2019-09-19
bare-metal: use ignition.config.url instead of flatcar.config.url

dongsupark:
Since upstream Container Linux changed the config URL to `ignition.config.url`, we should also replace `flatcar.config.url` with `ignition.config.url`.


invidian (approved):

dongsupark:
Oops. It was my mistake to close this PR.
Created the same one again: https://github.com/kinvolk/lokomotive-kubernetes/pull/83

PR #78, opened 2019-10-09
Bring back iPXE installation for Flatcar ARM images

dongsupark:
There is on-going work to support Flatcar Linux images for ARM64 architecture. (for example https://github.com/flatcar-linux/coreos-overlay/pull/88)
To make it integrated with Lokomotive, we should bring back the old iPXE installation code, because at the moment there is no way to natively install Flatcar ARM64 images on Packet.

This PR partly reverts https://github.com/kinvolk/lokomotive-kubernetes/pull/28.

**How to test:**

Add these variables to your arm64 workers/controllers:
```
ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/arm64-usr/packet.ipxe"
(controller_)type  = "c2.large.arm"
os_arch = "arm64"
os_channel = "alpha"
```
To test it for amd64 only the following variable is needed for workers/controllers:
```
ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/amd64-usr/packet.ipxe"
```
You can mix both types in controller and worker pools.

### DIFF
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -296,7 +296,8 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | controller_count | Number of controllers (i.e. masters) | 1 | 1 |
 | controller_type | Type of nodes to provision | "baermetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
-| os_version | Version of a Flatcar Linux release | current | 2191.5.0 |
+| os_arch    | Flatcar Container Linux architecture to install | amd64  | amd64, arm64 |
+| os_version | Version of a Flatcar Container Linux release, only for iPXE | current | 2191.5.0 |
 | ipxe_script_url | URL that contains iPXE script to boot Flatcar on the node over PXE | "" | https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/amd64-usr/packet.ipxe |
 | networking | Choice of networking provider | "calico" | "calico" or "flannel" |
 | network_mtu | CNI interface MTU (calico only) | 1480 | 8981 |
@@ -318,7 +319,8 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | type | Type of nodes to provision | "baremetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | labels | Comma separated labels to be added to the worker nodes | "" | "node.supernova.io/role=backend" |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
-| os_version | Version of a Flatcar Linux release | current | 2191.5.0 |
+| os_arch    | Flatcar Container Linux architecture to install | amd64  | amd64, arm64 |
+| os_version | Version of a Flatcar Container Linux release, only for iPXE | current | 2191.5.0 |
 | ipxe_script_url | URL that contains iPXE script to boot Flatcar on the node over PXE | "" | https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/amd64-usr/packet.ipxe |
 | cluster_domain_suffix | FQDN suffix for Kubernetes services answered by coredns. | "cluster.local" | "k8s.example.com" |
 | service_cidr | CIDR IPv4 range to assign to Kubernetes services | "10.3.0.0/16" | "10.3.0.0/24" |
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -296,6 +296,7 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | controller_count | Number of controllers (i.e. masters) | 1 | 1 |
 | controller_type | Type of nodes to provision | "baermetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
+| os_version | Version of a Flatcar Linux release | current | 2191.5.0 |
 | networking | Choice of networking provider | "calico" | "calico" or "flannel" |
 | network_mtu | CNI interface MTU (calico only) | 1480 | 8981 |
 | pod_cidr | CIDR IPv4 range to assign to Kubernetes pods | "10.2.0.0/16" | "10.22.0.0/16" |
@@ -316,6 +317,7 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | type | Type of nodes to provision | "baremetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | labels | Comma separated labels to be added to the worker nodes | "" | "node.supernova.io/role=backend" |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
+| os_version | Version of a Flatcar Linux release | current | 2191.5.0 |
 | cluster_domain_suffix | FQDN suffix for Kubernetes services answered by coredns. | "cluster.local" | "k8s.example.com" |
 | service_cidr | CIDR IPv4 range to assign to Kubernetes services | "10.3.0.0/16" | "10.3.0.0/24" |
 | setup_raid | Flag to create a RAID 0 from extra disks on a Packet node | "false" | "true" |
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -142,8 +142,6 @@ module "controller" {
   controller_count = 1
   controller_type  = "t1.small.x86"
 
-  ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/add-packet-ipxe/packet.ipxe"
-
   management_cidrs = [
     "0.0.0.0/0",       # Instances can be SSH-ed into from anywhere on the internet.
   ]
@@ -177,8 +175,6 @@ module "worker-pool-helium" {
 
   count = 2
   type  = "t1.small.x86"
-
-  ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/add-packet-ipxe/packet.ipxe"
 
   kubeconfig = "${module.controller.kubeconfig}"
 
@@ -301,7 +297,7 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | controller_type | Type of nodes to provision | "baermetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
 | os_version | Version of a Flatcar Linux release | current | 2191.5.0 |
-| ipxe_script_url | URL that contains iPXE script to boot Flatcar on the node over PXE | https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/add-packet-ipxe/packet.ipxe |
+| ipxe_script_url | URL that contains iPXE script to boot Flatcar on the node over PXE | "" | https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/amd64-usr/packet.ipxe |
 | networking | Choice of networking provider | "calico" | "calico" or "flannel" |
 | network_mtu | CNI interface MTU (calico only) | 1480 | 8981 |
 | pod_cidr | CIDR IPv4 range to assign to Kubernetes pods | "10.2.0.0/16" | "10.22.0.0/16" |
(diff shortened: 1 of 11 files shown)


pothos:
I'm trying out to restore support also for amd64 iPXE booting:

```
diff --git a/packet/flatcar-linux/kubernetes/controllers.tf b/packet/flatcar-linux/kubernetes/controllers.tf
index 0c41ac2..f66913a 100644
--- a/packet/flatcar-linux/kubernetes/controllers.tf
+++ b/packet/flatcar-linux/kubernetes/controllers.tf
@@ -41,7 +41,7 @@ resource "packet_device" "controllers" {
   hostname         = "${var.cluster_name}-controller-${count.index}"
   plan             = "${var.controller_type}"
   facilities       = ["${var.facility}"]
-  operating_system = "${var.os_arch == "arm64" ? "custom_ipxe" : format("flatcar_%s", var.os_channel)}"
+  operating_system = "${var.ipxe_script_url != "" ? "custom_ipxe" : format("flatcar_%s", var.os_channel)}"
   billing_cycle    = "hourly"
   project_id       = "${var.project_id}"
   user_data        = "${element(data.ct_config.controller-ignitions.*.rendered, count.index)}"
diff --git a/packet/flatcar-linux/kubernetes/workers/workers.tf b/packet/flatcar-linux/kubernetes/workers/workers.tf
index d92b0ad..747e8d8 100644
--- a/packet/flatcar-linux/kubernetes/workers/workers.tf
+++ b/packet/flatcar-linux/kubernetes/workers/workers.tf
@@ -3,7 +3,7 @@ resource "packet_device" "nodes" {
   hostname         = "${var.cluster_name}-${var.pool_name}-worker-${count.index}"
   plan             = "${var.type}"
   facilities       = ["${var.facility}"]
-  operating_system = "${var.os_arch == "arm64" ? "custom_ipxe" : format("flatcar_%s", var.os_channel)}"
+  operating_system = "${var.ipxe_script_url != "" ? "custom_ipxe" : format("flatcar_%s", var.os_channel)}"
   billing_cycle    = "hourly"
   project_id       = "${var.project_id}"
   ipxe_script_url  = "${var.ipxe_script_url}"
```


Edit: maybe the gpg key for ARM also should not be used for the installer then…
Edit2: Note: `/opt/installer` also can use the `-s` option for `flatcar-install` instead of a custom logic
Edit3: Result: I ended up in a ipxe booted instance but nothing got installed and the files supposed to be created by ignition where not present, e.g., `/opt/installer`.
Edit4: Ah, the OS is not installed… the wrong ignition template is used!
Edit5: pushed the above patch as commit

t-lo:
Adding @rata and @surajssd for review as this branch is in daily use for the Ampere benchmark project.
This branch allows for provisioning Arm clusters, which is currently not available from Packet. It also allows for provisioning hybrid clusters (i.e. mixed Arm / x86).

rata:
@t-lo let me know when it is ready for review. The first thing that needs to be cleaned up, from a quick look, is the commits and it's reverts. I don't see why we need a patch we later rebase. But I see this is a draft PR, so... let me know when it is ready for review or if I should look at something in particular :)

packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
```diff
@@ -0,0 +1,81 @@
...
+            local disk=$(lsblk -lnpd -I "$${major_numbers}" \
+              | sort -h -k 4,4 \
+              | head -1 \
+              | awk '{print $1}'
+            )
+
+            echo "$${disk}"
+          }
+
+          os_disk="$(select_install_disk $${major_numbers})"
```
rata: This ignores the `setup_raid` variable, so those features won't be available on arm without warning

pothos:
It uses the wrong Ignition file and therefore does not install to disk.
Also, the installation should use the `-s` flag for `flatcar-install`.

pothos:
I've switched to the installation ignition template and adapted it to the current needs of the alpha arm64 image.

pothos:
The current Flatcar alpha image does not have the `-s` option for `flatcar-install`, so I went with `-d /dev/sda` as temporary solution. All Packet arm64 servers just have just one SSD anyway.

packet/flatcar-linux/kubernetes/workers/cl/install.yaml.tmpl
pothos: The variable is passed to the second ignition and on first boot of the installed OS the RAID will be created as with the native Packet installation.

packet/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -22,4 +22,14 @@ module "bootkube" {
   enable_aggregation    = "${var.enable_aggregation}"
 
   certs_validity_period_hours = "${var.certs_validity_period_hours}"
+
+  container_images = {
+    calico           = "calico/node:v3.9.2"
```
invidian: Where those version come from?

docs/flatcar-linux/packet.md
```diff
@@ -316,6 +323,9 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | type | Type of nodes to provision | "baremetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | labels | Comma separated labels to be added to the worker nodes | "" | "node.supernova.io/role=backend" |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
+| os_arch    | Flatcar Container Linux architecture to install | amd64  | amd64, arm64 |
+| os_version | Version of a Flatcar Container Linux release | current | 2191.5.0 |
+| ipxe_script_url | URL that contains iPXE script to boot Flatcar on the node over PXE | https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/amd64-usr/packet.ipxe | https://scripts.foobar.com/ipxe-flatcar.ipxe |
```
invidian: I think there should be one more control variable called `use_ipxe`, so user don't have to explicitly unset that to use Packet images.

Or we could simply use `os_arch` for controlling when to use iPXE.

docs/flatcar-linux/packet.md
```diff
@@ -316,6 +323,9 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | type | Type of nodes to provision | "baremetal_0" | "t1.small.x86". See https://www.packet.com/developers/api/#plans for more |
 | labels | Comma separated labels to be added to the worker nodes | "" | "node.supernova.io/role=backend" |
 | os_channel | Flatcar Container Linux channel to install from | stable | stable, beta, alpha, edge |
+| os_arch    | Flatcar Container Linux architecture to install | amd64  | amd64, arm64 |
+| os_version | Version of a Flatcar Container Linux release | current | 2191.5.0 |
```
invidian: We should note, that this only works in iPXE mode.

docs/flatcar-linux/packet.md
invidian: Ah, also, it seems, that the documentation is wrong, as actual default value is `""`.

invidian (changes requested):
Mis-clicked the button in previous review. I found some nits.

Also, there seems to be several commits changing same fragment of code multiple times, it would be good to have those changes squashed before we merge. For example fixing "Flatcar Container Linux".

In general LGTM.

packet/flatcar-linux/kubernetes/bootkube.tf
pothos: Oh, current version should be `calico/*:v3.10.1`

invidian: Actually, I'm talking about all images here. Will we have to always sync it up now when we update bootkube module?

docs/flatcar-linux/packet.md
pothos: `os_arch` is independent from iPXE boot

docs/flatcar-linux/packet.md
pothos: Thanks, done

packet/flatcar-linux/kubernetes/bootkube.tf
pothos: We can also change the bootkube defaults to the values here.

pothos: Except for the `os_arch` suffix, this would need some more changes for bootkube.

invidian: Yeah, I think it would be better to propagate that to, I assume `terraform-render-bootkube` module, so we have single source of truth across our code.

pothos: Have changed it and will test now.

pothos:
TODO: Use `quay.io/kinvolk/calico-hostendpoint-controller:v0.0.2-ARCH`

kosyfrances:
I created a cluster with an arm worker and noticed that the calico-hostendpoint-controller pod was crashing. 
```
$ kubectl get pods -A                                                
NAMESPACE     NAME                                            READY   STATUS             RESTARTS   AGE
kube-system   calico-hostendpoint-controller-9f95569d-z5czn   0/1     CrashLoopBackOff   9          28m
kube-system   calico-node-nrhxp                               1/1     Running            0          26m
kube-system   calico-node-tfxhf                               1/1     Running            0          28m
....
```
This is not the case when you create a cluster with the previous defaults (i.e without arm).
The logs only give:
```
$ kubectl logs -n kube-system calico-hostendpoint-controller-9f95569d-z5czn
standard_init_linux.go:190: exec user process caused "exec format error"
```
My guess is that the container image for calico-hostendpoint-controller does not have the right architecture for arm machines.

packet/flatcar-linux/kubernetes/bootkube.tf
pothos: Works

pothos:
Fixed but with a temporary location in `kailueke/calico-hostendpoint-controller:v0.0.2`
Edit: Moved to `kinvolk/calico-hostendpoint-controller:v0.0.2`

```
NAMESPACE     NAME                                              READY   STATUS    RESTARTS   AGE     IP             NODE                  NOMINATED NODE   READINESS GATES
kube-system   calico-hostendpoint-controller-6f4c97c9db-7j9mq   1/1     Running   0          7m57s   10.2.184.1     k-test-he-worker-0    <none>           <none>
kube-system   calico-node-5brfr                                 1/1     Running   0          7m57s   10.88.72.141   k-test-controller-0   <none>           <none>
```
(k-test-he-worker-0 is c2.large.arm)

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -21,12 +21,24 @@ variable "project_id" {
...
 variable "os_channel" {
   type        = "string"
   default     = "stable"
   description = "Flatcar Container Linux channel to install from (stable, beta, alpha, edge)"
 }
 
+variable "os_version" {
+  type        = "string"
+  default     = "current"
+  description = "Flatcar Container Linux version to install (for example '2191.5.0' - see https://www.flatcar-linux.org/releases/)"
```
invidian: Here it should be also mentioned, that this is only for iPXE installation.

invidian (commented):
Codewise the PR looks good, however commits are a bit confusing. I didn't check if maybe GitHub is not messing up the order of the commits, I was reviewing it only via GitHub interface.

For the sake of possible debugging/reverting/bisecting in the future, I would re-organize it a bit, so we don't change things back and forth, as this intermediate step is not an intended state.

Here my suggestions for rebasing:

- 895faf3bf4788f9166dd40f0980f0512b33184f7 changes default value of `ipxe_script_url` to new URL, but documentation is changed to `""`. I think this commit should be dropped, as fda83988e886127005116d5d730e52ab62795424 changes it again later. I'd split and fixup to the first commit.

- 6055771ecbfdc5bdbcdd8bc8c118ad87c9d1bf53 adds temp key download. Then ed29e76c3ac7a49daeeb3f1e9c751038a3d2fe87 changes it again.

- 54f03253173d89af410dd203d4bced9835bfcaf5 adds `os_arch` variable and updates bootkube. Update should be perhaps done first in separated commit.

- fda83988e886127005116d5d730e52ab62795424 changes `operating_system` logic. Then it's changed again in d56eaae5f02803e01d4a7fa26ed72540a56598c1.

I would reorganize it to do the following:
- keep "Revert ..." commits as is at the beginning
- update bootkube (should be no-op)
- introduce `os_arch` variable and add everything related to it
- update reverted iPXE scripts, as they rely on `os_arch`

Not sure if I included all required steps, but I hope you get the idea.

On the other hand, if you don't want to spend time on it @pothos and other approve the PR, feel free to ignore me and just merge it.

kosyfrances (approved):
LGTM

packet/flatcar-linux/kubernetes/variables.tf
pothos: done

pothos:
Squashed some commits.

packet/flatcar-linux/kubernetes/variables.tf
```diff
@@ -30,7 +30,7 @@ variable "os_channel" {
 variable "os_version" {
   type        = "string"
   default     = "current"
-  description = "Flatcar Linux version to install (for example '2191.5.0' - see https://www.flatcar-linux.org/releases/)"
+  description = "Flatcar Container Linux version to install (for example '2191.5.0' - see https://www.flatcar-linux.org/releases/), only for iPXE"
```
invidian, line 39: Seems Flatcar Container Linux fix slipped into this commit.

invidian: Also `operating_system` changes doesn't seem to be related to documentation update.

packet/flatcar-linux/kubernetes/calico/host-endpoint-controller.yaml
```diff
@@ -27,7 +27,7 @@ spec:
     spec:
       serviceAccountName: calico-hostendpoint-controller
       containers:
-      - image: quay.io/kinvolk/calico-hostendpoint-controller:v0.0.2
+      - image: kinvolk/calico-hostendpoint-controller:v0.0.2
```
invidian, line 30: Shouldn't that be 0.0.3 by now?

invidian:
Also merged without waiting for CI :disappointed:

packet/flatcar-linux/kubernetes/calico/host-endpoint-controller.yaml
pothos: Comes through the other PR that is open.

PR #79, opened 2019-10-11
packet: fix issues when downloading rkt images

dongsupark:
Let's fix hidden issues discovered during tests for ARM support.

* rkt image name should have a prefix `docker://`
* rkt run should be running with stage1-fly
* rkt run/fetch should have an option `--insecure-options=image` to
  avoid intermittent cert errors.


iaguis:
Hm, I think fetching directly from `quay.io` (without `docker://`) should work, can you try again with your ARM images?

If we don't use `docker://` we should not pass `--insecure-options=image`. I'm a bit wary of turning off image signature checking, can you elaborate on what those intermittent cert errors are?

dongsupark:
Yeah, I would agree in general.
I cannot remember the error message.
Next week I'll test it again, and see if it really does not work.

dongsupark:
It turns out, the rkt fetch issue was regression from Upstream Go 1.12.9.
See also https://github.com/coreos/coreos-overlay/pull/3742
To make `rkt fetch` work again, Container Linux reverted the commit.
Flatcar stable - alpha should already include the bug fix. Edge will be soon released with the bug fix.

PR #83, opened 2019-10-31
bare-metal: use ignition.config.url instead of flatcar.config.url

dongsupark:
Since upstream Container Linux changed the config URL to `ignition.config.url`, we should also replace `flatcar.config.url` with `ignition.config.url`.

Replaces https://github.com/kinvolk/lokomotive-kubernetes/pull/70


invidian (approved):

rata (commented):
This looks good to me from a cursory look (for easier review, though, a reference to where this change was done upstream if it is easy to get would be great).

Is there a reason we are not merging this?

btw, what is the impact of not merging this? Is `flatcar.config.url` deprecated but still can be used or was it completely removed? If it was removed, I think is more urgent to merge this, probably :)

pothos:
Both are supported currently, but we should merge anyway.

PR #92, opened 2019-11-08
packet: Improve BGP support on workers

rata:
This patch series just improves the usage of BGP on worker nodes:

 * It is possible to disable BGP on worker pools that don't need it (default is false, so no backwards incompatible change is done)
 * There was an interpolation issue causing BGP sessions from all nodes to be deleted when a node was added/removed from the pool

The commits explain this in more detail.

### DIFF
--- a/packet/flatcar-linux/kubernetes/workers/workers.tf
+++ b/packet/flatcar-linux/kubernetes/workers/workers.tf
@@ -13,7 +13,7 @@ resource "packet_device" "nodes" {
 }
 
 resource "packet_bgp_session" "bgp" {
-  count          = "${var.count}"
+  count          = "${var.disable_bgp == "true" ? 0 : var.count}"
   device_id      = "${element(packet_device.nodes.*.id, count.index)}"
   address_family = "ipv4"
 }
--- a/packet/flatcar-linux/kubernetes/workers/workers.tf
+++ b/packet/flatcar-linux/kubernetes/workers/workers.tf
@@ -14,7 +14,7 @@ resource "packet_device" "nodes" {
 
 resource "packet_bgp_session" "bgp" {
   count          = "${var.disable_bgp == "true" ? 0 : var.count}"
-  device_id      = "${element(packet_device.nodes.*.id, count.index)}"
+  device_id      = "${packet_device.nodes.*.id[count.index]}"
   address_family = "ipv4"
 }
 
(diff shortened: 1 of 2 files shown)


packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -13,8 +13,8 @@ resource "packet_device" "nodes" {
 }
 
 resource "packet_bgp_session" "bgp" {
-  count          = "${var.count}"
-  device_id      = "${element(packet_device.nodes.*.id, count.index)}"
+  count          = "${var.disable_bgp == "true" ? 0 : var.count}"
+  device_id      = "${packet_device.nodes.*.id[count.index]}"
```
invidian, line 17: why not like this?
```suggestion
  device_id      = "${packet_device.nodes[count.index].id}"
```

rata: I currently have a mess of tf 0.11 and 0.12 syntax in my head, not sure if that will work. Are you sure that will have the same result?

Will try on Monday.

In any case, any reason to do it like that? Is something hidden (wouldn't be surprised if terraform has some odd behavior) there? Or what is the reason behind the suggestion? That it looks better?

invidian: My suggestion is just a more common pattern as far as I'm aware. It saves terraform time iterating over all resources and gathering the IDs, so you only select one in the end :smile: IMO it should work the same way.

rata: Hmmm... not sure it is more common (I copied the syntax from terraform-render-bootkube repo), and for sure **I** don't know how terraform parses it, it may take as well the same time. I think you are doing many assumptions in that statement. And, in any case, it is not interesting at all to improve that.

Also, it doesn't seem to be valid syntax for terraform 0.11 what you suggest. square bracket seems to be valid on lists, and packet_device.nodes is not a list (it possible to use square brackets on terraform 0.12, though, that changes), so you need to use the splat first (as it returns a list) and then index using square bracket.

Thanks a lot for the suggestion, but it seems it is not valid! (maybe you have a mess of tf 0.11 and 0.12 in your head too? :))

rata: Let me know if I'm missing something

rata:
@invidian PTAL. Is the very same thing, though :joy:

invidian (approved):
:see_no_evil:

rata:
@invidian thanks!

PR #95, opened 2019-11-19
[WIP] packet: create a symlink /dev/kmsg for starting kubelet

dongsupark:
Since k8s v1.15, kubelet cannot start without `/dev/kmsg` being available inside the container, because the OOM parser cannot open the file.

```
Failed to start OOM watcher open /dev/kmsg: no such file or directory
```

This happens only with flatcar-edge, where docker is used as container runtime in `kubelet-wrapper`.

See also https://github.com/kubernetes-sigs/kind/pull/664,
https://github.com/kubernetes/kubernetes/pull/74942

_not tested yet, please do not merge_
This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/114.


dongsupark:
Since Flatcar Edge 2345.99.1 started using rkt again, we do not need this PR any more.

PR #98, opened 2019-11-25
align "ssh_keys" variable in all providers

mauriciovasquezbernal:
This commit aligns all the supported providers to accept a list of public ssh
keys ("ssh_keys") to be copied to the nodes.

### DIFF
--- a/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
@@ -217,5 +217,4 @@ storage:
 passwd:
   users:
     - name: core
-      ssh_authorized_keys:
-        - "${ssh_authorized_key}"
+      ssh_authorized_keys: ${ssh_keys}
--- a/README.md
+++ b/README.md
@@ -47,7 +47,11 @@ module "aws-tempest" {
   dns_zone_id  = "Z3PAABBCFAKEC0"
 
   # configuration
-  ssh_authorized_key = "ssh-rsa AAAAB3Nz..."
+  ssh_keys = [
+    "ssh-rsa AAAAB3Nz...",
+    "ssh-rsa AAAAB3Nz...",
+  ]
+
   asset_dir          = "/home/user/.secrets/clusters/yavin"
 
   # optional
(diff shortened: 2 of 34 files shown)


surajssd (approved):
LGTM, just force push to run tests again. Packet is giving trouble again it seems.

alban (approved):
LGTM if you tested on packet + aws.

ipochi (approved):
lgtm

aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
```diff
@@ -217,5 +217,4 @@ storage:
 passwd:
   users:
     - name: core
-      ssh_authorized_keys:
-        - "${ssh_authorized_key}"
+      ssh_authorized_keys: ${ssh_keys}
```
johananl, line 220: Have we verified that interpolation works fine here both when specifying one key and when specifying multiple keys? I am concerned about the HCL->YAML conversion.

johananl (approved):
LGTM. Added one comment.

I think we should test this manually on any platform that isn't covered by CI (e.g. GCP).

aws/flatcar-linux/kubernetes/cl/controller.yaml.tmpl
mauriciovasquezbernal: I don't remember any test with multiple keys so far. I'll do it.
I'll try to test in another platform, probably azure.

johananl: I think we should test *all* platforms... Otherwise we may be breaking something on one platform.

alban: I expect it should work: the `ssh_keys` HCL variable is now a json string and json strings are valid YAML. But I have not tested :)

johananl: I expect it to work, too :-) But faith doesn't help improve software quality ;-)

mauriciovasquezbernal: I agree that the only way to be sure it's working would be to test on all platforms, but it's also true that the change is not platform dependent, if is working on packet so far it should work also on other platforms as well.

johananl: Yes. It's up to you. I've approved the PR anyway :-)

mauriciovasquezbernal: For the record: Tested with multiple keys in GCP and it works.

mauriciovasquezbernal:
When you are confident that your PR can't broke anything then a stupid mistake appears: https://github.com/kinvolk/lokomotive-kubernetes/pull/116/commits/afc4e928449d44cd1b3d5f7b27f09ef238a41685 (Part of https://github.com/kinvolk/lokomotive-kubernetes/pull/116). :man_facepalming:

PR #102, opened 2019-12-04
ci: Add Packet ARM cluster

pothos:
The cluster configuration is the same as for the existing
Packet cluster. The only difference is the server type
`c2.large.arm` (controller) and `c1.large.arm` (worker), the added variables `os_arch`/`os_channel`/`ipxe_script_url`,
and the worker count set to 1 instead of 2.


invidian (approved):
Nice! Looks good to me.

pothos:
Updated the iPXE URL and changed the worker type to `c1.large.arm` to save costs.

invidian (approved):
OK

PR #103, opened 2019-12-04
PR to incorporate changes w.r.t to bootkube supporting helm.

ipochi:
This PR updates the bootkube image to use the one which has helm patch

Refactors ssh.tf to copy calico hostendpoint manifests in calico helm chart.

### DIFF
--- a/packet/flatcar-linux/kubernetes/ssh.tf
+++ b/packet/flatcar-linux/kubernetes/ssh.tf
@@ -64,9 +64,8 @@ resource "null_resource" "copy-controller-secrets" {
   }
 }
 
-# Secure copy bootkube assets to ONE controller and start bootkube to perform
-# one-time self-hosted cluster bootstrapping.
-resource "null_resource" "bootkube-start" {
+# Secure copy bootkube assets to ONE controller.
+resource "null_resource" "copy-assets-dir" {
   depends_on = [
     module.bootkube,
     aws_route53_record.apiservers,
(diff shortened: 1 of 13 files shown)


pothos:
The multiarch image uploaded to quay may or may not work for arm64 because quay.io is currently rolling out support for that. That's why we were sticking to the `-ARCH` suffix, but currently no `-helm-arm64` and `-helm-amd64` are there, so I expect the CI to fail. Either verify that multiarch works on quay now and remove the `-ARCH` suffix or push the `-ARCH` tags as well.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -64,9 +64,8 @@ resource "null_resource" "copy-controller-secrets" {
   }
 }
 
-# Secure copy bootkube assets to ONE controller and start bootkube to perform
-# one-time self-hosted cluster bootstrapping.
-resource "null_resource" "bootkube-start" {
+# Secure copy bootkube assets to ONE controller
+resource "null_resource" "copy-assets-dir" {
```
invidian, line 68: In commit message:
>Due to the move to helm chart , there does not exist any directory named
  manifests or manifess-networking.

Typo + I'm don't fully understand this sentence. Could you elaborate @ipochi ?

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -64,9 +64,8 @@ resource "null_resource" "copy-controller-secrets" {
   }
 }
 
-# Secure copy bootkube assets to ONE controller and start bootkube to perform
-# one-time self-hosted cluster bootstrapping.
-resource "null_resource" "bootkube-start" {
+# Secure copy bootkube assets to ONE controller
```
invidian: Dot missing at the end of the comment.

packet/flatcar-linux/kubernetes/ssh.tf
```diff
@@ -84,15 +83,54 @@ resource "null_resource" "bootkube-start" {
...
+    type    = "ssh"
+    host    = "${packet_device.controllers.0.access_public_ipv4}"
+    user    = "core"
+    timeout = "15m"
+  }
 
   provisioner "file" {
     content     = data.template_file.host_protection_policy.rendered
-    destination = "$HOME/assets/manifests-networking/calico-policy.yaml"
+    destination = "$HOME/assets/charts/kube-system/calico/templates/calico-policy.yaml"
```
invidian, line 110: Can't we simply copy it to `manifests` directory? I'd rather not make it a part of the `calico` helm chart.

invidian (commented):
Just some nits, in general LGTM.

packet/flatcar-linux/kubernetes/ssh.tf
ipochi: > Can't we simply copy it to `manifests` directory? I'd rather not make it a part of the `calico` helm chart.

@invidian No we cannot  as calico-policy.yaml depends on CRDs installed with Calico.

Any manifests found in `manifests` directory get applied first before `charts` directory. Hence the installation fails for the `calico-host-endpoint controller`.

packet/flatcar-linux/kubernetes/ssh.tf
ipochi: > In commit message:
> 
> > Due to the move to helm chart , there does not exist any directory named
> > manifests or manifess-networking.
> 
> Typo + I'm don't fully understand this sentence. Could you elaborate @ipochi ?

reworded the commit message to reflect the reasons behind the change.
Please let me know if the reasons make sense now.

invidian: Yes, it makes more sense now. Thanks!

invidian (approved):
LGTM, we can merge when CI is :heavy_check_mark:

PR #106, opened 2019-12-06
docs: Document ARM support for Packet

pothos:
Add a section on ARM and hybrid clusters to the Packet docs.

### DIFF
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -239,6 +239,35 @@ supernova-helium-worker-1  Ready   node               10m  v1.14.1
 
 Learn about [maintenance](../topics/maintenance.md).
 
+## ARM Support and Hybrid Clusters
+
+Lokomotive and Flatcar support the Packet arm64 server types `c1.large.arm` and `c2.large.arm`.
+They can be used for both worker and controller nodes.
+Besides specifying them in `controller_type`/`type` you need to configure some additional variables
+in the respective controller/worker module:
+
+```
+os_arch = "arm64"
+os_channel = "alpha"
+ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/arm64-usr/packet.ipxe"
+```
+
+The `os_channel` variable is needed as long as the content of the iPXE script refers to `alpha.release…` as `base-url`.
+Adjust `os_channel` to the named channel there or remove it once the script uses `stable.release…` since `stable` is the default.
+The iPXE boot variable can be removed once Flatcar is available for installation in the Packet OS menu for the ARM servers.
+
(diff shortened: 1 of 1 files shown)


docs/flatcar-linux/packet.md
```diff
@@ -239,6 +239,35 @@ supernova-helium-worker-1  Ready   node               10m  v1.14.1
...
+Besides specifying them in `controller_type`/`type` you need to configure some additional variables
+in the respective controller/worker module:
+
+```
+os_arch = "arm64"
+os_channel = "alpha"
+ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/arm64-usr/packet.ipxe"
+```
+
+The channel variable is needed as long as the content of the iPXE script refers to `alpha.release…`.
```
kosyfrances: channel variable or `os_channel` variable?

docs/flatcar-linux/packet.md
```diff
@@ -239,6 +239,35 @@ supernova-helium-worker-1  Ready   node               10m  v1.14.1
...
+Adjust it to the named channel there or remove it once the script uses `stable.release…`.
```
kosyfrances: Is it when the URL has that string or can you please paste an example?

docs/flatcar-linux/packet.md
pothos: thanks

docs/flatcar-linux/packet.md
pothos: Tried to make it more clear by saying that it's the value for `base-url`.

docs/flatcar-linux/packet.md
```diff
@@ -239,6 +239,35 @@ supernova-helium-worker-1  Ready   node               10m  v1.14.1
...
+os_channel = "alpha"
+ipxe_script_url = "https://raw.githubusercontent.com/kinvolk/flatcar-ipxe-scripts/arm64-usr/packet.ipxe"
+```
+
+The `os_channel` variable is needed as long as the content of the iPXE script refers to `alpha.release…` as `base-url`.
+Adjust `os_channel` to the named channel there or remove it once the script uses `stable.release…` since `stable` is the default.
+The iPXE boot variable can be removed once Flatcar is available for installation in the Packet OS menu for the ARM servers.
+
+If you have a hybrid cluster with both x86 and ARM nodes, you need to either use Docker multiarch images such as the standard
+`debian:latest` or `python:latest` images, or restrict the Pods to the nodes of the correct architecture with an entry like
```
kosyfrances, line 260: `pods` -->  lower case

pothos: Pods as a Kubernetes concept, so capital

kosyfrances (approved):
LGTM

PR #107, opened 2019-12-09
packet: Remove temporary ARM signature key

pothos:
Since Flatcar Container Linux now has official
ARM64 releases, the developer signature workaround
is not needed anymore and hinders accepting offically
signed images.


dongsupark (commented):
From a quick look, it looks good.
Though let's merge it a little later, when we released a new Edge successfully.
It will be done in a couple of days.

pothos:
Currently booting alpha images is broken, so I would like to merge now.

dongsupark (approved):
I see

PR #109, opened 2019-12-09
packet/flatcar-linux: add support for CLC snippets

mpolednik:
Both kubernetes and kubernetes/workers module in Packet lack the
ability to specify CLC (Container Linux Configuration) snippets that
are used to modify the instances (or devices in Packet).

We can solve that by following the pattern set by code in other
cloud providers, only adapting it to Packet's structure. The variable
names remain consistent.

Signed-off-by: Martin Polednik <m.polednik@gmail.com>

### DIFF
--- a/docs/advanced/customization.md
+++ b/docs/advanced/customization.md
@@ -65,7 +65,7 @@ View the Container Linux Config [format](https://coreos.com/os/docs/1576.4.0/con
 
 Write Container Linux Configs *snippets* as files in the repository where you keep Terraform configs for clusters (perhaps in a `clc` or `snippets` subdirectory). You may organize snippets in multiple files as desired, provided they are each valid.
 
-[AWS](../flatcar-linux/aws.md#cluster) and [Azure](../flatcar-linux/azure.md#cluster) clusters allow populating a list of `controller_clc_snippets` or `worker_clc_snippets`.
+[AWS](../flatcar-linux/aws.md#cluster), [Azure](../flatcar-linux/azure.md#cluster) and [Packet](../flatcar-linux/packet.md#cluster) clusters allow populating a list of `controller_clc_snippets` or `worker_clc_snippets`.
 
 ```
 module "aws-nemo" {
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -258,6 +258,7 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 | facility | Packet Region in which the instance(s) should be deployed | https://www.packet.com/developers/api/#facilities. Eg: "ams1" |
 | management_cidrs | List of CIDRs to allow SSH access to the nodes | ["153.79.80.1/16", "59.60.10.1/32"] |
 | node_private_cidr | Private CIDR obtained from Packet for the project and facility | 10.128.16.32/25 |
+| controller_clc_snippets | Controller Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 #### Worker module
 
@@ -269,6 +270,7 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 | facility | Packet Region in which the instance(s) should be deployed | https://www.packet.com/developers/api/#facilities. Eg: "ams1" |
 | pool_name | Name of the worker pool. Used in setting hostname | "helium" |
 | kubeconfig | Kubeconfig to be used in worker pools | "${module.controller.kubeconfig} |
+| clc_snippets | Worker Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 #### DNS Zone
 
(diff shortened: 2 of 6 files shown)


packet/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -28,6 +28,12 @@ variable "type" {
   description = "Packet instance type for workers, can be changed afterwards to recreate the nodes"
 }
 
+variable "worker_clc_snippets" {
```
rata: What about using just clc_snippets, to match the variable name on AWS? https://github.com/kinvolk/lokomotive-kubernetes/blob/17043dcb736c070df1cb136a1f84ddb6af378cff/aws/flatcar-linux/kubernetes/workers/variables.tf#L73-L77

docs/advanced/customization.md
```diff
@@ -65,7 +65,7 @@ View the Container Linux Config [format](https://coreos.com/os/docs/1576.4.0/con
 
 Write Container Linux Configs *snippets* as files in the repository where you keep Terraform configs for clusters (perhaps in a `clc` or `snippets` subdirectory). You may organize snippets in multiple files as desired, provided they are each valid.
 
-[AWS](../flatcar-linux/aws.md#cluster) and [Azure](../flatcar-linux/azure.md#cluster) clusters allow populating a list of `controller_clc_snippets` or `worker_clc_snippets`.
+[AWS](../flatcar-linux/aws.md#cluster), [Azure](../flatcar-linux/azure.md#cluster) and [Packet](../flatcar-linux/packet.md#cluster) clusters allow populating a list of `controller_clc_snippets` or `worker_clc_snippets`.
```
rata, line 68: Oh, I realize we don't have a `workers.tf` in `packet/flatcar-linux/kubernetes/` so that very simple example will not work on Packet. That probably happened when moving to support worker pools. I can fill a bug for that... :)

Let's add Packet to that list as this PR does, and we can fix the example for Packet in another PR/issue (to keep the scope of this reduced, I understand you don't have much time). Will that work for you?

docs/flatcar-linux/packet.md
```diff
@@ -269,6 +270,7 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 | facility | Packet Region in which the instance(s) should be deployed | https://www.packet.com/developers/api/#facilities. Eg: "ams1" |
 | pool_name | Name of the worker pool. Used in setting hostname | "helium" |
 | kubeconfig | Kubeconfig to be used in worker pools | "${module.controller.kubeconfig} |
+| worker_clc_snippets | Worker Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
```
rata: Rename to `clc_snippets`

rata (commented):
From a quick look, LGTM. Thanks a lot for the patch! :-)

Some small comments

rata (commented):
Some other things to take into account:

* The ignition config file in AWS has a max size of 16kB (https://github.com/kinvolk/lokomotive-kubernetes/commit/48f7a727c5c9798fab26bceb99774e61d7065b30) and there might be a similar limitation on Packet. Check that using an ignition config of size >= 16kB works on packet. If it doesn't, document the limitation. 

* Check if there is documentation available on the max user_data size an instance in Packet can use. If there is, please document/link that in the terraform variable description.

If you, @mpolednik, lack the time to check those, let me know :-)

docs/flatcar-linux/packet.md
rata: I think that should be the name *for the worker pool module* (packet/flatcar-linux/kubernetes/workers)


On the top level module (packet/flatcar-linux/kubernetes) it should be with the workers prefix, but we don't support that for packet (seems like a bug to me) and these docs are about the worker module. So, dropping the worker prefix here is the right thing, I think :)

mauriciovasquezbernal:
@rata I am already checking the points you mentioned, will be back with an answer soon.

mpolednik:
Pushed s/workers_// (also re-verified that it works).

For the docs/structure, an issue to make Packet consistent with other providers seems like reasonable way to me.

rata (approved):
@mpolednik thank you for the changes and testing it! LGTM :)

Can do the other things (and even a test for the ci would be great) as follow-up, IMHO. I'd like to wait for @mauriciovasquezbernal (he is in a different timezone) and have his review too.

mauriciovasquezbernal:
LGTM, thanks @mpolednik for this PR.

I can confirm that a user-data > 16kB works, however I haven't been able to find if there is a limit for it. We can update docs later on when this find this.

mauriciovasquezbernal (approved):

mauriciovasquezbernal:
@mholttech I forgot to mention, although it is possible to pass an ignition configuration bigger than 16kB we have noticed that the provision time increases a lot when using sizes bigger than 64kB, in some cases the deployment even timeouts. Once we have an answer from packet I'll update the documentation.

PR #110, opened 2019-12-12
docs/packet: document CLC snippets limitation

mauriciovasquezbernal:
- move CLC snippets to optional variables.
- document issue when using big snippets.

### DIFF
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -287,7 +287,6 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 | facility | Packet Region in which the instance(s) should be deployed | https://www.packet.com/developers/api/#facilities. Eg: "ams1" |
 | management_cidrs | List of CIDRs to allow SSH access to the nodes | ["153.79.80.1/16", "59.60.10.1/32"] |
 | node_private_cidr | Private CIDR obtained from Packet for the project and facility | 10.128.16.32/25 |
-| controller_clc_snippets | Controller Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 #### Worker module
 
@@ -299,7 +298,6 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 | facility | Packet Region in which the instance(s) should be deployed | https://www.packet.com/developers/api/#facilities. Eg: "ams1" |
 | pool_name | Name of the worker pool. Used in setting hostname | "helium" |
 | kubeconfig | Kubeconfig to be used in worker pools | "${module.controller.kubeconfig} |
-| clc_snippets | Worker Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 #### DNS Zone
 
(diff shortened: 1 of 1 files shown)


docs/flatcar-linux/packet.md
```diff
@@ -362,9 +361,15 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
...
 | reservation_ids | Map Packet hardware reservation IDs to instances. | {} | { worker-0 = "55555f20-a1fb-55bd-1e11-11af11d11111" } |
 | reservation_ids_default | Default hardware reservation ID for nodes not listed in the `reservation_ids` map. | "" | "next-available"|
+| clc_snippets [[1]](#clc-snippets-limitation) | Worker Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 Documentation about Packet hardware reservation id can be found here: https://support.packet.com/kb/articles/reserved-hardware.
 
+#### CLC Snippets Limitation
+
+The size of the CLC snippets affects the time a node needs to deploy in a severe way, for instance a snippet of 64kB increases the deployment time by 5 minutes and a snippet of 128kB could cause the deployment to timeout.
+It is not clear for us yet what is causing this behaviour, we are investigating it: https://github.com/kinvolk/lokomotive-kubernetes/issues/111.
```
invidian: Maybe we could re-phrase to:
```suggestion
See [issue #111](https://github.com/kinvolk/lokomotive-kubernetes/issues/111) for more details.
```

docs/flatcar-linux/packet.md
```diff
@@ -362,9 +361,15 @@ Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`
 | taints | Comma separated list of custom taints for all workers in the worker pool | "" | "clusterType=staging:NoSchedule,nodeType=storage:NoSchedule" |
 | reservation_ids | Map Packet hardware reservation IDs to instances. | {} | { worker-0 = "55555f20-a1fb-55bd-1e11-11af11d11111" } |
 | reservation_ids_default | Default hardware reservation ID for nodes not listed in the `reservation_ids` map. | "" | "next-available"|
+| clc_snippets [[1]](#clc-snippets-limitation) | Worker Container Linux Config snippets | [] | [example](../advanced/customization.md#usage) |
 
 Documentation about Packet hardware reservation id can be found here: https://support.packet.com/kb/articles/reserved-hardware.
 
+#### CLC Snippets Limitation
+
+The size of the CLC snippets affects the time a node needs to deploy in a severe way, for instance a snippet of 64kB increases the deployment time by 5 minutes and a snippet of 128kB could cause the deployment to timeout.
```
invidian: Hm, it's not the CLC Snippets limitation, but user-data size. I think we should mention here, that Lokomotive already consumes X kB from this 64kB.

I would also mention, that on other platforms the user-data size is limited to 16 kB (so user is aware, that one snippet may not necessarily work on other platforms).

Lastly, I think it would be nice to suggest here a way to check the size of the snippet. Maybe some Terraform code fragment should be placed here? Or maybe Lokomotive should expose the snippet size as an output, so user can measure it's size?

invidian (commented):
Left some suggestions.

docs/flatcar-linux/packet.md
mauriciovasquezbernal: I'll rephrase that a little bit indicating that the problem is with the user-data and that CLC snippets are added there.

I haven't been able to find an easy way to get the size of user-data without having to expose that as an output.

I tried the following but the reported size is not right becase there are a lot of empty spaces.

```
terraform apply -target=module.controller.data.ct_config.controller-ignitions
terraform state show module.controller.ct_config.controller-ignitions | grep rendered | wc -c | numfmt --to=iec
```

Do you have any suggestion?

mauriciovasquezbernal: Btw, I think a bigger work is needed to document limitations in other platforms as well, for instance the 16kB that you mentioned for AWS and so on.

rata (approved):
Nice catch with the var documented as required. And thanks a lot for this improvement! LGTM =)

invidian (approved):

PR #114, opened 2019-12-12
docs: Rephrase the current state of arm64 until it's stable

pothos:
Rephrase the support of arm64 until the Flatcar stable arm64 release is done, official Packet installation is there, and all components support arm64.


invidian (approved):
LGTM

PR #116, opened 2019-12-17
Fix azure 

mauriciovasquezbernal:
This PR fixes the azure module by solving three issues that broke it. More details in the commits.

### DIFF
--- a/azure/flatcar-linux/kubernetes/controllers.tf
+++ b/azure/flatcar-linux/kubernetes/controllers.tf
@@ -115,7 +115,7 @@ resource "azurerm_network_interface" "controllers" {
 resource "azurerm_network_interface_backend_address_pool_association" "controllers" {
   count = "${var.controller_count}"
 
-  network_interface_id    = "${azurerm_network_interface.controllers.*.id}"
+  network_interface_id    = "${element(azurerm_network_interface.controllers.*.id, count.index)}"
   ip_configuration_name   = "ip0"
   backend_address_pool_id = "${azurerm_lb_backend_address_pool.controller.id}"
 }
--- a/azure/flatcar-linux/kubernetes/controllers.tf
+++ b/azure/flatcar-linux/kubernetes/controllers.tf
@@ -70,13 +70,14 @@ resource "azurerm_virtual_machine" "controllers" {
     custom_data    = "${element(data.ct_config.controller-ignitions.*.rendered, count.index)}"
   }
 
-  # Azure mandates setting an ssh_key, even though Ignition custom_data handles it too
+  # Azure mandates setting an ssh_key, provide just a single key as the
+  # others are handled with Ignition custom_data.
   os_profile_linux_config {
     disable_password_authentication = true
 
     ssh_keys {
       path     = "/home/core/.ssh/authorized_keys"
-      key_data = "${var.ssh_keys}"
+      key_data = "${element(var.ssh_keys, 0)}"
     }
   }
 
--- a/azure/flatcar-linux/kubernetes/workers/workers.tf
+++ b/azure/flatcar-linux/kubernetes/workers/workers.tf
@@ -42,13 +42,14 @@ resource "azurerm_virtual_machine_scale_set" "workers" {
     custom_data          = "${data.ct_config.worker-ignition.rendered}"
   }
 
-  # Azure mandates setting an ssh_key, even though Ignition custom_data handles it too
+  # Azure mandates setting an ssh_key, provide just a single key as the
+  # others are handled with Ignition custom_data.
   os_profile_linux_config {
     disable_password_authentication = true
 
     ssh_keys {
       path     = "/home/core/.ssh/authorized_keys"
-      key_data = "${var.ssh_authorized_key}"
+      key_data = "${element(var.ssh_keys, 0)}"
     }
   }
 
--- a/azure/flatcar-linux/kubernetes/workers/workers.tf
+++ b/azure/flatcar-linux/kubernetes/workers/workers.tf
@@ -73,7 +73,11 @@ resource "azurerm_virtual_machine_scale_set" "workers" {
 
   # eviction policy may only be set when priority is Low
   priority        = "${var.priority}"
-  eviction_policy = "${var.priority == "Low" ? "Delete" : null}"
+  # It is not possible to assign an optional argument in terraform 0.11,
+  # (https://github.com/hashicorp/terraform/issues/17968)
+  # however azurerm provider v1.35 ignores the value if the priority
+  # is not low. https://github.com/terraform-providers/terraform-provider-azurerm/blob/3b0ba54174ed56d2082228349b1bc19c38b75eca/azurerm/resource_arm_virtual_machine_scale_set.go#L866
+  eviction_policy = "${var.priority == "Low" ? "Delete" : "Deallocate" }"
 }
 
 # Scale up or down to maintain desired number, tolerating deallocations.


azure/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -115,7 +115,7 @@ resource "azurerm_network_interface" "controllers" {
 resource "azurerm_network_interface_backend_address_pool_association" "controllers" {
   count = "${var.controller_count}"
 
-  network_interface_id    = "${azurerm_network_interface.controllers.*.id}"
+  network_interface_id    = "${element(azurerm_network_interface.controllers.*.id, count.index)}"
```
rata, line 119: Just curious: why `element()` and not just using square brackets?

azure/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -73,7 +73,11 @@ resource "azurerm_virtual_machine_scale_set" "workers" {
 
   # eviction policy may only be set when priority is Low
   priority        = "${var.priority}"
-  eviction_policy = "${var.priority == "Low" ? "Delete" : null}"
+  # It is not possible to assign an optional argument in terraform 0.11,
+  # (https://github.com/hashicorp/terraform/issues/17968)
+  # however azurerm provider v1.35 ignores the value if the priority
+  # is not low. https://github.com/terraform-providers/terraform-provider-azurerm/blob/3b0ba54174ed56d2082228349b1bc19c38b75eca/azurerm/resource_arm_virtual_machine_scale_set.go#L866
```
rata, line 80: Uff, you had to look at the code to catch this? Awesome work!

This can be set to "null" on terraform 0.12, right? In that case, would it make sense to add an "XXX: When upgrading to terraform 0.12 consider to simplify this by doing..."? If that makes sense to you, as you are actively working on the 0.12 PR to... whatever you feel more comfortable with, but removing this by using `null` in tf 0.12 seems way nicer :)

azure/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -70,13 +70,14 @@ resource "azurerm_virtual_machine" "controllers" {
...
-  # Azure mandates setting an ssh_key, even though Ignition custom_data handles it too
+  # Azure mandates setting an ssh_key, provide just a single key as the
+  # others are handled with Ignition custom_data
   os_profile_linux_config {
     disable_password_authentication = true
 
     ssh_keys {
       path     = "/home/core/.ssh/authorized_keys"
-      key_data = "${var.ssh_keys}"
+      key_data = "${element(var.ssh_keys, 0)}"
```
rata, line 80: Sorry to insist on this, but I'm not sure I understand when this is used. Is this used at all? Or as the first boot happens with ignition, this is overwritten on first boot? Not sure the server state if ignition fails for some reason.

I think being very clear about this might help: either use `key_data = "unused see comment"` (or state in the comment that this is set but **never** used in practice) if it is not used **at all in any case**, or try to use all the keys if they are indeed used, isn't that possible?

Also, if this happens to be used in some cases, can't we just assign all the keys? Something like `key_data = "${join("\n",var.ssh_keys)}"` doesn't work?

rata (approved):
LGTM, added some small comments but can be improved later if you prefer (we can create issues for them, if you consider any worth improving later)

azure/flatcar-linux/kubernetes/controllers.tf
mauriciovasquezbernal: This PR is against current master branch that uses Terraform 0.11.

azure/flatcar-linux/kubernetes/workers/workers.tf
mauriciovasquezbernal: > Uff, you had to look at the code to catch this? Awesome work!
>

Yes, I had fun looking at the code!

> This can be set to "null" on terraform 0.12, right?

Right, in-fact that's the problem, it was updated to use `null` but we are still running Terrraform 0.11. Actually the PR for Terraform 0.12 will have to revert this change as the current solution is good for it. Since I'm the one working on the other PR I'll avoid making a comment :), I hope TF0.12 will be merged soon too.

azure/flatcar-linux/kubernetes/controllers.tf
mauriciovasquezbernal: > Sorry to insist on this, but I'm not sure I understand when this is used. Is this used at all? Or as the first boot happens with ignition, this is overwritten on first boot? Not sure the server state if ignition fails for some reason.

The azure provider requires at least one ssh key when `disable_password_authentication` is `true`. The confusion could  be that there is not any similar requirement in AWS, GCE or Packet and we have to do that to fulfill the requirement.

In this case the key is copied twice to the server, one time by Azure (no idea how) and the other time by ignition.

> I think being very clear about this might help: either use `key_data = "unused see comment"` (or state in the comment that this is set but **never** used in practice) if it is not used **at all in any case**, or try to use all the keys if they are indeed used, isn't that possible?
>

I could have try to use multiple keys, but given that they are already handled by ignition I avoided that. I did something like a minimal-effort approach to fulfill the requirement of having a least one key.

> Also, if this happens to be used in some cases, can't we just assign all the keys? Something like `key_data = "${join("\n",var.ssh_keys)}"` doesn't work?

I did something similar in https://github.com/kinvolk/lokomotive-kubernetes/pull/98, then I found it wasn't working and, as already mentioned, I implemented the fasted solution I found.

If you still think it's a good idea to try to copy all the keys here I can give it a try.

azure/flatcar-linux/kubernetes/controllers.tf
```diff
@@ -70,13 +70,14 @@ resource "azurerm_virtual_machine" "controllers" {
     custom_data    = "${element(data.ct_config.controller-ignitions.*.rendered, count.index)}"
   }
 
-  # Azure mandates setting an ssh_key, even though Ignition custom_data handles it too
+  # Azure mandates setting an ssh_key, provide just a single key as the
+  # others are handled with Ignition custom_data
```
invidian: Dot missing at the end of the sentence.

azure/flatcar-linux/kubernetes/controllers.tf
invidian, line 80: I think if you join the keys with newline here, it should work just fine. Also, they should be put in `/home/core/.ssh/authorized_keys.d`, so `update-ssh-keys` does not override them.

azure/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -42,13 +42,14 @@ resource "azurerm_virtual_machine_scale_set" "workers" {
     custom_data          = "${data.ct_config.worker-ignition.rendered}"
   }
 
-  # Azure mandates setting an ssh_key, even though Ignition custom_data handles it too
+  # Azure mandates setting an ssh_key, provide just a single key as the
+  # others are handled with Ignition custom_data
```
invidian: Ditto dot here.

invidian (changes requested):
Just some nits, nice work @mauriciovasquezbernal

azure/flatcar-linux/kubernetes/controllers.tf
mauriciovasquezbernal: Ignition keys are also saved to `~/.ssh/authorized_keys`, any reason why these should be placed in a different file?

invidian: Hmm, I'm not sure then.

mauriciovasquezbernal: Nevermind, we cannot put them in another place: 
https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#ssh_keys
"NOTE: Due to a limitation in the Azure VM Agent the only allowed path is /home/{username}/.ssh/authorized_keys."

invidian (approved):

azure/flatcar-linux/kubernetes/workers/workers.tf
invidian: Still missing?

azure/flatcar-linux/kubernetes/controllers.tf
rata: Terraform 0.11 supports `[]`. But element is fine, too. (just has some problems with dependencies)

azure/flatcar-linux/kubernetes/workers/workers.tf
rata: :ok_hand:

azure/flatcar-linux/kubernetes/controllers.tf
rata: > The azure provider requires at least one ssh key when `disable_password_authentication` is `true`. The confusion could be that there is not any similar requirement in AWS, GCE or Packet and we have to do that to fulfill the requirement.
> 
> In this case the key is copied twice to the server, one time by Azure (no idea how) and the other time by ignition.

No, sorry I wasn't clear. I mean what happens for example if we set an invalid ssh key? What are the downsides of doing that?

Or if they are downsides, why not use all the keys? (using a " \n" join as I mentioned https://github.com/kinvolk/lokomotive-kubernetes/pull/116#discussion_r361979875 and @invidian here https://github.com/kinvolk/lokomotive-kubernetes/pull/116#discussion_r362429469)

rata: > I did something similar in #98, then I found it wasn't working and, as already mentioned, I implemented the fasted solution I found.

Oh, sorry, just read this. Ok, if it's not working, I'd like to understand if we need this. Like, what is the downside of using an invalid key, something like `key_data = "ssh-rsa ignoreme"`?

azure/flatcar-linux/kubernetes/workers/workers.tf
mauriciovasquezbernal: I haven't pushed yet because was trying do so something about the ssh_keys :P.

azure/flatcar-linux/kubernetes/controllers.tf
mauriciovasquezbernal: Just for the record. I tried to copy all keys here too but it's not possible to do in Terraform 0.11.

The configuration would look like:

```
os_profile_linux_config {
    disable_password_authentication = true
    ssh_keys {
      path     = "/home/core/.ssh/authorized_keys"
      key_data = "${element(var.ssh_keys, 0)}"
    }
    ssh_keys {
      path     = "/home/core/.ssh/authorized_keys"
      key_data = "${element(var.ssh_keys, 1)}"
    }

  # other ssh_keys blocks...
  }
```

It is not possible to generate all those blocks depending on the length of `var.ssh_keys`.

rata: I suggested using join, though, not that :) (https://github.com/kinvolk/lokomotive-kubernetes/pull/116#discussion_r361979875)

mauriciovasquezbernal: That was not working either. The only thing  I didn't try was to add a fake ssh key (key_data = "ssh-rsa ignoreme").

rata: Sorry, thanks! :)

PR #137, opened 2020-01-15
packet: Abstract DNS records

pothos:
Currently an AWS account is needed to deploy a cluster in Packet because the DNS provider is harcoded to be Route53. This is a limitation as some users do not have an AWS account.

This PR proposes a solution to that by abstracting the DNS provider concept. Now the user can choose which provider to use by defining a module like the following in their configuration:

```
module "dns" {
  source = "/home/mvb/lokomotive-kubernetes/dns/route53/"

  entries = module.controller.dns_entries
  aws_zone_id = "Z1QEP94KB3YNR3"
}
```

It is even possible to no use a DNS provider at all and create the DNS entries by hand by following this procedure:

```bash
  # Create controller nodes (you could also create worker nodes to save time)
  terraform apply -target=module.controller.packet_device.controllers

  # Get list of DNS entries to be created
  terraform output dns_entries

  # Create the DNS entries by hand

  # Finish deploying the cluster
  terraform apply
```

This PR creates a DNS provider module for Route53, more modules can be easily added in the future.

### DIFF
--- a/dns/route53/main.tf
+++ b/dns/route53/main.tf
@@ -1,0 +1,27 @@
+variable "entries" {
+  type = list(
+    object({
+      name    = string
+      type    = string
+      ttl     = number
+      records = list(string)
+    })
+  )
+}
+
+variable "aws_zone_id" {
+  type        = string
+  description = "AWS Route53 DNS Zone ID (e.g. Z3PAABBCFAKEC0)"
+}
+
+resource "aws_route53_record" "dns-records" {
+  count = length(var.entries)
+
+  # Route53 DNS Zone where record should be created
+  zone_id = var.aws_zone_id
+
+  name    = var.entries[count.index].name
+  type    = var.entries[count.index].type
+  ttl     = var.entries[count.index].ttl
+  records = var.entries[count.index].records
+}
--- a/dns/route53/require.tf
+++ b/dns/route53/require.tf
@@ -1,0 +1,9 @@
+# Terraform version and plugin versions
+
+terraform {
+  required_version = ">= 0.12.0"
+
+  required_providers {
+    aws = "2.31.0"
+  }
+}
--- a/docs/dns/route53.md
+++ b/docs/dns/route53.md
@@ -1,0 +1,43 @@
+# Route53 DNS provider
+
+This module uses the AWS Route53 service to setup the DNS entries required for the cluster.
+
+Include the following snippet in your configuration to use this module:
+
+```tf
+module "dns" {
+  source = "git::https://github.com/kinvolk/lokomotive-kubernetes//dns/route53?ref=<hash>"
+
+  entries = module.controller.dns_entries
+
+  aws_zone_id = "<zone_id>" # e.g. Z3PAABBCFAKEC0
+}
+```
+
+## AWS Credentials
+
+Login to your AWS IAM dashboard and find your IAM user. Select "Security Credentials" and create an access key. Save the id and secret to a file that can be referenced in configs.
+
+```
+[default]
+aws_access_key_id = xxx
+aws_secret_access_key = yyy
+```
+
+!!! tip
+    Other standard AWS authentication methods can be used instead of specifying `shared_credentials_file` under the provider's config. For more information see the [docs](https://www.terraform.io/docs/providers/aws/#authentication).
+
+Configure the AWS provider to use your access key credentials in a `providers.tf` file.
+
+```
+provider "aws" {
+  version = "2.31.0"
+  alias   = "default"
+
+  region                  = "eu-central-1"
+  shared_credentials_file = "/home/user/.config/aws/credentials"
+}
+```
+
+Additional configuration options are described in the `aws` provider [docs](https://www.terraform.io/docs/providers/aws/).
+
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -1,6 +1,6 @@
 # Packet
 
-In this tutorial, we'll create a Kubernetes cluster on [Packet](https://packet.net) with [Flatcar Container Linux](https://www.flatcar-linux.org/). For external DNS, [Route53](https://aws.amazon.com/route53/) will be used.
+In this tutorial, we'll create a Kubernetes cluster on [Packet](https://packet.net) with [Flatcar Container Linux](https://www.flatcar-linux.org/).
 
 We'll declare a Kubernetes cluster using the Lokomotive Terraform module. Then apply the changes to create DNS records, controller machines, worker machines and TLS assets.
 
@@ -9,8 +9,7 @@ Controllers are provisioned to run an `etcd-member` peer and a `kubelet` service
 ## Requirements
 
 * Packet account, Project ID and [API key](https://support.packet.com/kb/articles/api-integrations) (Note, that the term "Auth Token" is also used to refer to the API key in the packet docs)
-* AWS Account and IAM credentials
-* AWS Route53 DNS Zone (registered Domain Name or delegated subdomain)
+* [DNS Zone](#dns-zone)
 * Terraform v0.12.x and [terraform-provider-ct](https://github.com/poseidon/terraform-provider-ct) installed locally
 
 ## Terraform Setup
@@ -38,10 +37,6 @@ cd infra/clusters
 
 ## Provider
 
-### AWS
-
-Login to your AWS IAM dashboard and find your IAM user. Select "Security Credentials" and create an access key. Save the id and secret to a file that can be referenced in configs.
-
 ```
 [default]
 aws_access_key_id = xxx
@@ -62,10 +57,9 @@ provider "aws" {
 }
 ```
 
-Additional configuration options are described in the `aws` provider [docs](https://www.terraform.io/docs/providers/aws/).
+
 
 !!! tip
-    AWS regions are listed in [docs](http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region) or with `aws ec2 describe-regions`.
     The Packet facilities (i.e. data centers) list can be dynamically queried using the [API docs](https://www.packet.com/developers/api/#facilities).
 
 ### Packet
@@ -88,9 +82,8 @@ locals {
 module "controller" {
   source = "git::https://github.com/kinvolk/lokomotive-kubernetes//packet/flatcar-linux/kubernetes?ref=<hash>"
 
-  # Route53
-  dns_zone    = "packet.example.com"
-  dns_zone_id = "Z3PAABBCFAKEC0"
+  # DNS configuration
+  dns_zone    = "myclusters.example.com"
 
   # configuration
   ssh_keys = [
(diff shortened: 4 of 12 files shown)


packet/flatcar-linux/kubernetes/dns.tf
```diff
@@ -0,0 +1,34 @@
+module "dns-entries" {
```
invidian: We could instantiate both `route53` and `manual` and have a knob for which DNS provider should be used. Something like
```
entries = var.dns_provider == "route53" ? local.dns_entries : 0
```

This way user does not have to modify the module, which I believe should  be discouraged.

The downside is that at some point we will have more modules, which will be empty, but this is kind of Terraform limitation... But I think then we can take this outside of kubernetes module and let user define them and drive them using `-target`.

invidian (commented):
Just one comment, otherwise LGTM.

packet/flatcar-linux/kubernetes/dns.tf
pothos: Then route53 would require to get some dummy credentials?

invidian: Oh, it seems `aws` provider does not lazily evaluate the credentials :(

```
provider "aws" {
  version = "2.31.0"

  region = "us-east-1"
}

resource "aws_route53_record" "dns-records" {
  count = 0

  type = "A"
  zone_id = "nonexistent"
  name = ""
}
```

Following code requires AWS credentials :man_facepalming:

mauriciovasquezbernal: Yes, unfortunately it's a problem I haven't been able to find a feasible solution. There are two Terraform limitations that make this impossible to implement in a clean way now:
1. `module` does not support the count parameter: https://github.com/hashicorp/terraform/issues/953.
2. It's not possible to use a variable in the source of a module https://github.com/hashicorp/terraform/issues/1439.

packet/flatcar-linux/kubernetes/bootkube.tf
```diff
@@ -6,7 +6,7 @@ module "bootkube" {
   # Cannot use cyclic dependencies on controllers or their DNS records
   api_servers          = [format("%s-private.%s", var.cluster_name, var.dns_zone)]
   api_servers_external = [format("%s.%s", var.cluster_name, var.dns_zone)]
-  etcd_servers         = aws_route53_record.etcds.*.fqdn
+  etcd_servers         = [for device in packet_device.controllers: format("%s-etcd%d.%s.", var.cluster_name, index(packet_device.controllers, device), var.dns_zone)]
```
invidian: Note, that this breaks the implicit dependency of bootkube module on `aws_route53_record.etdcs` resource. I think explicit dependency should be added then.

invidian (changes requested):

packet/flatcar-linux/kubernetes/dns.tf
invidian:
```
variable "use_aws" {
	default = 0
}

provider "aws" {
  version = "2.31.0"

  region = var.use_aws == 0 ? "nonexisting" : "us-east-1"
  skip_region_validation = var.use_aws == 0 ? true : false
  skip_credentials_validation = var.use_aws == 0 ? true : false
  skip_requesting_account_id = var.use_aws == 0 ? true : false
  skip_metadata_api_check = var.use_aws == 0 ? true : false
  skip_get_ec2_platforms = var.use_aws == 0 ? true : false
  access_key = var.use_aws == 0 ? "nonexistent" : null
  secret_key = var.use_aws == 0 ? "nonexistent" : null
}

resource "aws_route53_record" "dns-records" {
  count = var.use_aws

  type = "A"
  zone_id = "nonexistent"
  name = ""
}
```

Seems to work :see_no_evil: Not sure if all fields I set are needed though.

invidian: Note, to test that, one can use `docker run -it --entrypoint /bin/sh --rm hashicorp/terraform`.

dns/route53/main.tf
```diff
@@ -0,0 +1,27 @@
+variable "entries" {
```
invidian, line 1: This module must define what version of Terraform and AWS provider is required to use it.

invidian (changes requested):

dns/route53/require.tf
```diff
@@ -0,0 +1,9 @@
+# Terraform version and plugin versions
+
+terraform {
+  required_version = ">= 0.12.0"
+}
+
+provider "aws" {
+  version = "2.31.0"
```
invidian: Please use `required_providers` syntax to avoid instantiating the provider. See https://www.terraform.io/docs/configuration/modules.html#provider-version-constraints-in-modules for more details.

docs/flatcar-linux/packet.md
```diff
@@ -307,20 +292,35 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
...
+Clusters create few DNS A records to resolve to controller instances. For instance `${cluster_name}.${dns_zone}` is used by workers and `kubectl` to access the apiserver(s). In this example, the cluster's apiserver would be accessible at `tempest.myclusters.example.com`.
 
-You'll need a registered domain name or delegated subdomain on AWS Route53. You can set this up once and create many clusters with unique names.
+In order to create such DNS entries you'll need a registered domain/subdomain and to define a DNS module that takes as input the DNS entries required by the controller module.
 
 ```tf
-resource "aws_route53_zone" "zone-for-clusters" {
-  name = "aws.example.com."
+module "dns" {
+  source = "git::https://github.com/kinvolk/lokomotive-kubernetes//dns/providerx?ref=<hash>"
```
pothos: Please write `<provider>` instead of `providerx`.

docs/flatcar-linux/packet.md
```diff
@@ -159,6 +132,19 @@ module "controller" {
   node_private_cidr = "10.128.156.0/25"
 }
 
+# DNS module that creates the required DNS entries for the cluster.
+# More details in DNS Zone section of this document.
+module "dns" {
+  source = "git::https://github.com/kinvolk/lokomotive-kubernetes//dns/route53?ref=<hash>"
```
invidian, line 119: Those examples should be copy-pastable for the user, even if they point to some old commit hash or to the master branch.

invidian (changes requested):

docs/flatcar-linux/packet.md
pothos: Currently `<hash>` is used, so it's okay to keep it consistent for now.

docs/flatcar-linux/packet.md
mauriciovasquezbernal: Done.

dns/route53/require.tf
mauriciovasquezbernal: Ok, done!

dns/route53/main.tf
mauriciovasquezbernal: Done.

docs/dns/route53.md
```diff
@@ -0,0 +1,47 @@
+# Route53 DNS provider
+
+This module uses the AWS Route53 service to setup the DNS entries required for the cluster.
+
+Include the following snip in your configuration to use this module:
```
invidian:
```suggestion
Include the following snippet in your configuration to use this module:
```

docs/flatcar-linux/packet.md
```diff
@@ -219,9 +205,9 @@ Apply the changes to create the cluster.
 ```sh
 $ terraform apply
 ...
-module.aws-tempest.null_resource.bootkube-start: Still creating... (4m50s elapsed)
-module.aws-tempest.null_resource.bootkube-start: Still creating... (5m0s elapsed)
-module.aws-tempest.null_resource.bootkube-start: Creation complete after 11m8s (ID: 3961816482286168143)
+module.packet-tempest.null_resource.bootkube-start: Still creating... (4m50s elapsed)
```
invidian:
```suggestion
module.controller.null_resource.bootkube-start: Still creating... (4m50s elapsed)
```

docs/flatcar-linux/packet.md
```diff
@@ -307,20 +292,35 @@ Check the [variables.tf](https://github.com/kinvolk/lokomotive-kubernetes/blob/m
 
 #### DNS Zone
 
-Clusters create a DNS A record `${cluster_name}.${dns_zone}` to resolve a network load balancer backed by controller instances. This FQDN is used by workers and `kubectl` to access the apiserver(s). In this example, the cluster's apiserver would be accessible at `tempest.aws.example.com`.
+Clusters create few DNS A records to resolve to controller instances. For instance `${cluster_name}.${dns_zone}` is used by workers and `kubectl` to access the apiserver(s). In this example, the cluster's apiserver would be accessible at `tempest.myclusters.example.com`.
```
invidian:
```suggestion
Clusters create few DNS A records to resolve to controller instances. For example `${cluster_name}.${dns_zone}` is used by workers and `kubectl` to access the apiserver(s). In this example, the cluster's apiserver would be accessible at `tempest.myclusters.example.com`.
```

`instance` => `example` to avoid using word `instance(s)` twice.

packet/flatcar-linux/kubernetes/bootkube.tf
invidian: We discussed, that bootkube itself should wait long enough before the records become available, so it should be fine, so long user does not specify `-parallelism=1`. I think with this, it will break.

packet/flatcar-linux/kubernetes/outputs.tf
```diff
@@ -5,3 +5,34 @@ output "kubeconfig-admin" {
...
   value = module.bootkube.kubeconfig-kubelet
 }
+
+output "dns_entries" {
+  value = concat(
+    # etcd
+    [
+      for device in packet_device.controllers:
+      {
+        name    = format("%s-etcd%d.%s.", var.cluster_name, index(packet_device.controllers, device), var.dns_zone),
```
invidian: Can we please move `format` calls to either `locals` or to `null_resource`, so they are not duplicated?

invidian (changes requested):
LGTM, just left few nit comments.

packet/flatcar-linux/kubernetes/outputs.tf
mauriciovasquezbernal: Oh yeah, I didn't want to use a null_resource but locals is a good solution :).

packet/flatcar-linux/kubernetes/outputs.tf
```diff
@@ -5,3 +5,34 @@ output "kubeconfig-admin" {
 output "kubeconfig" {
   value = module.bootkube.kubeconfig-kubelet
 }
+
+output "dns_entries" {
+  value = concat(
+    # etcd
+    [
+      for device in packet_device.controllers:
```
invidian, line 13: You could just do `for index, device in packet_device.controllers` IIRC. Then there is no need to call a function :wink:

invidian (approved):
LGTM

packet/flatcar-linux/kubernetes/outputs.tf
mauriciovasquezbernal: Thanks!

invidian:
@mauriciovasquezbernal I suggest rebasing to latest master, so kube-hunter on Packet is not flaky.

invidian (approved):

invidian:
@mauriciovasquezbernal can you please rebase?

mauriciovasquezbernal:
@invidian done!

invidian (approved):

surajssd (approved):
LGTM

rata:
Very nice work, thanks! :)

PR #138, opened 2020-01-16
packet/flatcar-linux: Added support for MetalLB + Calico (+ Packet BGP)

DGollings:
I've created a plugin (module) that easily adds a MetalLB LoadBalancer type to the cluster. For easy exposing of services.

Should work near out of the box with minimal configuration/changes to the original scripts


chrskhl:
Hi @DGollings. Thanks for the PR. As part of the first major release of Lokomotive we'll be updating the repo with major changes in the coming weeks. One of those additions will be a metallb component. It will also introduce major changes to the way terraform is exposed.

Sorry the opaqueness. As part of this upcoming initial release, we'll be moving all work and planning into the open source project which will greatly improve visibility into the project.

Thus, I'm going to close this.

DGollings:
shame, was quite a lot of effort, but it was educational so not a loss. Looking forward to the official version then.

rata:
@DGollings thanks a lot for the effort, though! I hope we can focus effort (if it makes sense to you and others) on one metallb thing for Packet and incorporate the learnings :)

DGollings:
I agree, make sure to check out the code and in particular the patches as it's a drop in module for your current calico configuration (as far as I've been able to tell)

Would be nice if there's a road map somewhere so I don't spend effort on other shortcomings I've encountered (IMHO). Because I definitely like the idea of flatcar Linux :)

rata:
I'll check the code and glad you like Flatcar Container Linux! =)

Sorry again, but there will be a public roadmap :-D

PR #139, opened 2020-01-16
docs: Fix asset path for Packet documentation

BrainBlasted:
In the "Verify" step of the Packet docs, the asset path is actually the one from the tempest documentation rather than the one defined in the example configuration above.

This commit changes the directory to the correct one.


invidian (approved):
LGTM

PR #140, opened 2020-01-17
Simplify handling providers

invidian:
As a prerequisite for simplified version of #137.


aws/flatcar-linux/kubernetes/require.tf
```diff
@@ -2,24 +2,12 @@
...
-}
-
-provider "template" {
-  version = "~> 2.1"
-}
 
-provider "tls" {
-  version = "~> 2.0"
+  required_providers {
+    aws      = ">= 2.31.0"
```
rata: Doesn't this allow version 3.0.0? I don't think we want that, as that would mean the aws provider probably broke compatibility. Why not maintain the symbols used before?

invidian: Oops, bad copy-paste. will fix.

invidian: Fixed.

aws/flatcar-linux/kubernetes/require.tf
```diff
@@ -2,24 +2,12 @@
...
-}
 
-provider "tls" {
-  version = "~> 2.0"
+  required_providers {
+    aws      = "2.31.0"
+    local    = "~> 1.2"
+    null     = "~> 2.1"
+    template = "~> 2.1"
+    tls      = "~> 2.0"
```
rata, line 12: no ct provider needs to be added here? (I see you added it on bare-metal)

packet/flatcar-linux/kubernetes/require.tf
```diff
@@ -2,32 +2,14 @@
...
-}
-
-provider "packet" {
-  version = "~> 2.7.3"
-}
 
-provider "aws" {
-  version = "2.31.0"
+  required_providers {
+    aws      = "2.31.0"
```
rata: I the think previous (deleted) missing `~>` was probably a bug introduced recently (tf 0.12 migration?)

packet/flatcar-linux/kubernetes/workers/require.tf
```diff
@@ -2,24 +2,14 @@
...
-}
-
-provider "tls" {
-  version = "~> 2.0"
-}
 
-provider "packet" {
-  version = "~> 2.7.3"
+  required_providers {
+    aws      = "2.31.0"
```
rata: idem

rata (approved):
LGTM.

There are some small things that can be fixed while we are there , I think, but is reasonable in any case. If you prefer not to fix the comments and makes sense to address them for you too, please create an issue for them :)

aws/flatcar-linux/kubernetes/require.tf
invidian: It wasn't there, but yeah, it should be added. Adding in a separate commit.

packet/flatcar-linux/kubernetes/require.tf
invidian: Could be, but I'd say it's out of scope for this PR to fix that?

invidian: Created #143 for it.

packet/flatcar-linux/kubernetes/workers/require.tf
invidian: Created #143 for it.

packet/flatcar-linux/kubernetes/require.tf
rata: :ok_hand:

invidian:
Updated:
- I split changes per platform, so they are easier for review.
- Checked all platforms for missing constraints and added them where needed.
- Make sure docs are updated as well.

packet/flatcar-linux/kubernetes/require.tf
mauriciovasquezbernal: > I the think previous (deleted) missing `~>` was probably a bug introduced recently (tf 0.12 migration?)

It wasn't me :stuck_out_tongue_closed_eyes:

surajssd (approved):
LGTM, this has reduced a lot of unnecessary boilerplate \o/

pothos (approved):

invidian:
Will rebase and merge after #146.

PR #141, opened 2020-01-17
packet: make Route53 dependency optional

invidian:
Depends on #140 
Replaces #137 

TODO:
- modify docs to mention, that Route53 dependency is optional. If so, cluster must be created in 2 steps, using `terraform apply -target module.controller.null_resource.dns_entries`.
- improve commit message


invidian:
Closing in favor of #137.

PR #144, opened 2020-01-20
Add kubelet DS

surajssd:


invidian (commented):
Hm, what about moving labels and taints to .env file for other platforms? Shouldn't it be done for all of them?

Otherwise LGMT.

surajssd:
@invidian done.

invidian (approved):

PR #146, opened 2020-01-20
scripts/kube-hunter-manifests/config.yaml: enable debug output and switch to patched image

### LINKED ISSUE #145: kube-hunter CI job is flaky
It sometimes doesn't finish within 7 minutes for some reason, which makes CI job to fail. We should investigate that.

invidian:
To mitigate the issues found while investigating #145 and to be prepared for further debugging, when some other issue occurs.

Closes #145

### DIFF
--- a/scripts/kube-hunter-manifests/config.yaml
+++ b/scripts/kube-hunter-manifests/config.yaml
@@ -29,7 +29,7 @@ spec:
       serviceAccountName: kube-hunter
       containers:
       - name: kube-hunter
-        image: aquasec/kube-hunter
+        image: quay.io/kinvolk/kube-hunter:b0c6610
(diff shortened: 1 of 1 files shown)


pothos:
In no CI case did it print out the report anymore.

invidian:
>In no CI case did it print out the report anymore.

Nice catch, thanks for checking that! It seem reverting https://github.com/kinvolk/kube-hunter/commit/a4a8c71653132bda0bc751fecb346854df0552a0 was not a good idea :smile: 

I dropped the `--log DEBUG` patch for now then and it should be fine now :crossed_fingers:

invidian:
The output is now visible in CI.

pothos (approved):

PR #152, opened 2020-01-24
packet: fix DNS entries output

mauriciovasquezbernal:
Create a null_resource to force the name of the domains to be computed,
otherwise they will be null on the output.

Also fix the documentation for the user to get the list of entries from the
null resource as the output from internal modules is nost saved on the state
on Terraform 0.12.

### DIFF
--- a/packet/flatcar-linux/kubernetes/outputs.tf
+++ b/packet/flatcar-linux/kubernetes/outputs.tf
@@ -7,32 +7,5 @@ output "kubeconfig" {
 }
 
 output "dns_entries" {
-  value = concat(
-    # etcd
-    [
-      for device in packet_device.controllers:
-      {
-        name    = local.etcd_fqdn[index(packet_device.controllers, device)],
-        type    = "A",
-        ttl     = 300,
-        records = [device.access_private_ipv4],
-      }
-    ],
-    [
-      # apiserver public
-      {
-        name    = local.api_external_fqdn
-        type    = "A",
-        ttl     = 300,
-        records = packet_device.controllers.*.access_public_ipv4,
-      },
-      # apiserver private
-      {
-        name    = local.api_fqdn,
-        type    = "A",
-        ttl     = 300,
-        records = packet_device.controllers.*.access_private_ipv4,
-      },
-    ]
-  )
+  value = local.dns_entries
 }
--- a/docs/flatcar-linux/packet.md
+++ b/docs/flatcar-linux/packet.md
@@ -281,10 +281,10 @@ Lokomotive implements support for some [DNS providers](../dns/), if your provide
 
 ```bash
   # Create controller nodes (you could also create worker nodes to save time)
-  terraform apply -target=module.controller.packet_device.controllers
+  terraform apply -target=module.controller.null_resource.dns_entries
 
   # Get list of DNS entries to be created
-  terraform output dns_entries
+  terraform state show module.controller.null_resource.dns_entries
 
   # Create the DNS entries by hand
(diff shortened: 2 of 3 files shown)


packet/flatcar-linux/kubernetes/outputs.tf
```diff
@@ -7,32 +7,5 @@ output "kubeconfig" {
...
-      # apiserver private
-      {
-        name    = local.api_fqdn,
-        type    = "A",
-        ttl     = 300,
-        records = packet_device.controllers.*.access_private_ipv4,
-      },
-    ]
-  )
+  value = local.dns_entries
```
invidian, line 10: Maybe this should point to `null_resource.dns_entries.triggers.value` ?

invidian: Then there is no need to use `local.dns_entries` and those computation can be put directly into `null_resource`?

mauriciovasquezbernal: I think it'll not work because triggers must be a map of strings and the DNS module expects this output to follow a specific structure: https://github.com/kinvolk/lokomotive-kubernetes/blob/c3088e9c853c5ea723bc64ce2901b730965929d4/dns/route53/main.tf#L1-L10.

invidian: What about:
```hcl
resource "null_resource" "foo" {
  triggers = {
    value = jsonencode([
      {
        name    = "foo",
        type    = "A",
        ttl     = 300,
        records = ["8.8.8.8"],
      },
    ])
  }
}

output "dns_entries" {
  value = jsondecode(null_resource.foo.triggers.value)
}
```

That seems to work fine for me:
```
0 ✓ (12.0s) 10:00:05 invidian@dellxps15mateusz /tmp/test $ terraform apply

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # null_resource.foo will be created
  + resource "null_resource" "foo" {
      + id       = (known after apply)
      + triggers = {
          + "value" = jsonencode(
                [
                  + {
                      + name    = "foo"
                      + records = [
                          + "8.8.8.8",
                        ]
                      + ttl     = 300
                      + type    = "A"
                    },
                ]
            )
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

null_resource.foo: Creating...
null_resource.foo: Creation complete after 0s [id=3127116993682709981]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

Outputs:

dns_entries = [
  {
    "name" = "foo"
    "records" = [
      "8.8.8.8",
    ]
    "ttl" = 300
    "type" = "A"
  },
]
```

mauriciovasquezbernal: I see. I'm not sure if that is better than currently we have in this PR, I cannot see what's wrong with having a local defining some data structure and then having a `null_resource` forcing them to be created. The fact that the entries in the null resource are encoded in json and then they have to be decode in the output is not that clean IHMO.

If you still think current approach is not good I could change it to keep things rolling :).

invidian: I'm also not the fan of `jsondecode`, so I think locals are fine then :smile:

invidian (approved):
LGTM

PR #155, opened 2020-01-29
aws: split variable os_image into os_name, os_channel, os_version

alban:
_(description updated to match the discussion in the PR)_

Before this patch, the variable os_image was composite in the form
'$os-$channel' such as 'flatcar-stable'.

This patch changes split the variable to follow the model in Packet.
We now have:
- os_name = coreos | flatcar
- os_channel = stable | beta | alpha | edge (like in Packet)
- os_version = current | 2261.99.0 (new feature, like in Packet)

### DIFF
--- a/aws/flatcar-linux/kubernetes/variables.tf
+++ b/aws/flatcar-linux/kubernetes/variables.tf
@@ -41,10 +41,22 @@ variable "worker_type" {
   description = "EC2 instance type for workers"
 }
 
-variable "os_image" {
+variable "os_name" {
   type        = string
-  default     = "flatcar-stable"
-  description = "AMI channel for a CoreOS Container Linux derivative (coreos-stable, coreos-beta, coreos-alpha, flatcar-stable, flatcar-beta, flatcar-alpha)"
+  default     = "flatcar"
+  description = "Name of Operating System (coreos or flatcar)"
+}
+
+variable "os_channel" {
+  type        = string
+  default     = "stable"
+  description = "AMI channel for the OS (stable, beta, alpha, edge)"
+}
+
+variable "os_version" {
+  type        = string
+  default     = "current"
+  description = "Version of the OS (current or numeric version such as 2261.99.0)"
 }
 
 variable "disk_size" {
--- a/aws/flatcar-linux/kubernetes/workers/ami.tf
+++ b/aws/flatcar-linux/kubernetes/workers/ami.tf
@@ -1,11 +1,10 @@
 locals {
   # Pick a CoreOS Container Linux derivative
-  # coreos-stable -> CoreOS Container Linux AMI
-  # flatcar-stable -> Flatcar Container Linux AMI
   ami_id = local.flavor == "flatcar" ? data.aws_ami.flatcar.image_id : data.aws_ami.coreos.image_id
 
-  flavor  = split("-", var.os_image)[0]
-  channel = split("-", var.os_image)[1]
+  flavor  = var.os_name
+  channel = var.os_channel
+  ver = var.os_version == "current" ? "" : var.os_version
 }
 
 data "aws_ami" "coreos" {
@@ -24,7 +23,7 @@ data "aws_ami" "coreos" {
 
   filter {
     name   = "name"
-    values = ["CoreOS-${local.flavor == "coreos" ? local.channel : "stable"}-*"]
+    values = ["CoreOS-${local.flavor == "coreos" ? local.channel : "stable"}-${local.flavor == "coreos" ? local.ver : ""}*"]
   }
 }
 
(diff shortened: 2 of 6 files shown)


alban:
I tested this with the following `albandemoedge.lokocfg`:
```
variable "asset_dir" {
	type = "string"
}

variable "ssh_pubkeys" {
	type = "list"
}

cluster "aws" {
	asset_dir = "${pathexpand(var.asset_dir)}"
	cluster_name = "albandemoedge"
	os_image = "flatcar-edge-2261.99.0"
	dns_zone = "..."
	dns_zone_id = "..."
	ssh_pubkeys = var.ssh_pubkeys
}

component "contour" {}
```

invidian (commented):
@alban thanks for the PR.

I'm thinking, maybe instead of splitting `os_image`, we could align those parameters with what we currently have for Packet: https://github.com/kinvolk/lokomotive-kubernetes/blob/master/packet/flatcar-linux/kubernetes/variables.tf#L25-L31. Then we provide better consistency across the providers.

So instead `os_image`, we would have `os_channel` and `os_version`.

`os_version` could be `current` by default, so then we map it to `*` for AMI.

What do you think?

alban:
That sounds ok to me. Note that the splitting of `os_image` existed before my PR though, so that would be a breaking change.

But we still need to specify whether we pick CoreOS or Flatcar. So on AWS we would have 3 variables:
- os_name = coreos | flatcar
- os_channel = edge
- os_version = current | 2261.99.0

On Packet, we will not have `os` but we currently have support for `os_arch`, so there will still be differences of support between AWS and Packet.

alban:
Note that I think we should have a way to overwrite the AMI directly for CI purposes, but that's out of scope of this PR and that would be a change in a different place, in controllers.tf:
```
resource "aws_instance" "controllers" {
  ami       = local.ami_id
```

aws/flatcar-linux/kubernetes/variables.tf
```diff
@@ -41,10 +41,22 @@ variable "worker_type" {
...
-  default     = "flatcar-stable"
-  description = "AMI channel for a CoreOS Container Linux derivative (coreos-stable, coreos-beta, coreos-alpha, flatcar-stable, flatcar-beta, flatcar-alpha)"
+  default     = "flatcar"
+  description = "Name of Operating System (coreos or flatcar)"
+}
+
+variable "os_channel" {
+  type        = string
+  default     = "stable"
+  description = "AMI channel for the OS (stable, beta, alpha, edge"
```
invidian:
```suggestion
  description = "AMI channel for the OS (stable, beta, alpha, edge)"
```

aws/flatcar-linux/kubernetes/workers/variables.tf
```diff
@@ -34,10 +34,22 @@ variable "instance_type" {
...
-  default     = "flatcar-stable"
-  description = "AMI channel for a CoreOS Container Linux derivative (coreos-stable, coreos-beta, coreos-alpha, flatcar-stable, flatcar-beta, flatcar-alpha)"
+  default     = "flatcar"
+  description = "Name of Operating System (coreos or flatcar)"
+}
+
+variable "os_channel" {
+  type        = string
+  default     = "stable"
+  description = "AMI channel for the OS (stable, beta, alpha, edge"
```
invidian:
```suggestion
  description = "AMI channel for the OS (stable, beta, alpha, edge)"
```

aws/flatcar-linux/kubernetes/workers/ami.tf
```diff
@@ -24,7 +23,7 @@ data "aws_ami" "coreos" {
 
   filter {
     name   = "name"
-    values = ["CoreOS-${local.flavor == "coreos" ? local.channel : "stable"}-*"]
+    values = ["CoreOS-${local.flavor == "coreos" ? local.channel : "stable"}-${local.flavor == "coreos" ? local.ver : ""}*"]
```
invidian, line 26: I'm not sure about this logic here. Why is it needed? why can't we do `ver = var.os_version == "current" ? "*" : var.os_version` in `locals` and just use `local.ver` directly here?

invidian (changes requested):
Just some small nits, otherwise LGTM

aws/flatcar-linux/kubernetes/workers/ami.tf
alban: It's for the same reason we check the flavor before using the channel, and defaulting to stable: to avoid Terraform failing because no 'edge' AMI can be found in the 'coreos' images.

Both terraform "aws_ami" sections (flatcar and coreos) are always evaluated even if only one will end up being used.

When the user selects os_name `flatcar` and os_version `2345.99.1`, that version exists on AWS for 'flatcar' but not for 'coreos', and that would make terraform fail if we didn't guard it with this check.

invidian: Ah, that make sense. We could possibly add `count` to both AMIs, to disable calculating the another one to simplify all that, but that seems like out of scope for this PR.

Thanks for explanation :)

invidian (dismissed):
LGTM

invidian (commented):
Actually, also documentation should be updated to reflect those changes: https://github.com/kinvolk/lokomotive-kubernetes/blob/master/docs/flatcar-linux/aws.md#optional

alban:
@invidian doc updated

invidian (approved):

PR #165, opened 2020-02-06
aws: remove dependency on coreos images

alban:
The end-of-life announcement for CoreOS Container Linux
(https://coreos.com/os/eol/) says that:

    > On or after September 1, [...] OS images will be removed from
    > AWS, Azure, and Google Compute Engine

Previously, the "aws_ami coreos" section was always evaluated, even if
the user selects flatcar. The CoreOS images were not used though. If the
CoreOS AMIs were deleted, that would break the deployment of Flatcar.

This patch removes this dependency using the terraform code pattern with
resources count = 0. See also:
https://github.com/hashicorp/terraform/issues/16726#issuecomment-346037079

-----

TODO:
- [x] Tests

### DIFF
--- a/aws/flatcar-linux/kubernetes/ami.tf
+++ b/aws/flatcar-linux/kubernetes/ami.tf
@@ -1,6 +1,6 @@
 locals {
-  # Pick a CoreOS Container Linux derivative
-  ami_id = local.flavor == "flatcar" ? data.aws_ami.flatcar.image_id : data.aws_ami.coreos.image_id
+  # Pick a Container Linux derivative
+  ami_id = "${element(concat(data.aws_ami.flatcar.*.image_id, data.aws_ami.coreos.*.image_id, list("")), 0)}"
 
   flavor  = var.os_name
   channel = var.os_channel
@@ -8,6 +8,8 @@ locals {
 }
 
 data "aws_ami" "coreos" {
+  count  = "${local.flavor == "coreos" ? 1 : 0}"
+
   most_recent = true
   owners      = ["595879546273"]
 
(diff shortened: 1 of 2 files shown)


aws/flatcar-linux/kubernetes/ami.tf
```diff
@@ -1,13 +1,15 @@
 locals {
-  # Pick a CoreOS Container Linux derivative
-  ami_id = local.flavor == "flatcar" ? data.aws_ami.flatcar.image_id : data.aws_ami.coreos.image_id
+  # Pick a Container Linux derivative
+  ami_id = local.flavor == ${element(concat(data.aws_ami.flatcar.*.image_id, data.aws_ami.coreos.*.image_id, list("")), 0)}
```
invidian: `local.flavor == ` part should be dropped here? Currently `ami_id` will be bool if I understand the code correctly.

alban: Yes, my mistake. I'm fixing it and I will test again.

invidian (dismissed):
LGTM

alban:
Thanks for the quick review :)

I'd like to do more manual test before merging. Then, I will merge it.

alban:
@invidian I updated the PR: I initially forgot the file `workers/ami.tf`. Could you review it again?

invidian (approved):

aws/flatcar-linux/kubernetes/ami.tf
```diff
@@ -1,13 +1,15 @@
 locals {
-  # Pick a CoreOS Container Linux derivative
-  ami_id = local.flavor == "flatcar" ? data.aws_ami.flatcar.image_id : data.aws_ami.coreos.image_id
+  # Pick a Container Linux derivative
+  ami_id = "${element(concat(data.aws_ami.flatcar.*.image_id, data.aws_ami.coreos.*.image_id, list("")), 0)}"
```
mauriciovasquezbernal, line 3: Why to use string interpolation if that is not needed in Terraform 0.12? :sob:

invidian: Oops :man_facepalming: Nice catch @mauriciovasquezbernal!

alban: I am not aware of the change in Terraform 0.12. How would you write this line instead?

invidian: You just need to remove quotes and interpolation Syntax and you just leave pure expression.

PR #166, opened 2020-02-07
bootkube/resources/charts/kubelet: make /sys RW

invidian:
As stated as requirement in
kubernetes/kubernetes#43704 (comment).

Also, bootstrap-kubelet also use writeable /sys from host, as this is
provided by the rkt stage1: https://coreos.com/rkt/docs/latest/running-fly-stage1.html#how-does-it-work

### DIFF
--- a/bootkube/resources/charts/kubelet/templates/kubelet-ds.yaml
+++ b/bootkube/resources/charts/kubelet/templates/kubelet-ds.yaml
@@ -73,7 +73,6 @@ spec:
           mountPath: /run
         - name: sys
           mountPath: /sys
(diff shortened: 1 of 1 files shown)


rata:
@invidian Can you explain in the commit message why doing that should fix the error (and link to relevant comment, etc.)?

invidian:
>@invidian Can you explain in the commit message why doing that should fix the error (and link to relevant comment, etc.)?

I changed the description, as this actually does not resolve the problem I've tried to address. However, this patch is still something we should have.

surajssd (approved):
LGTM

PR #167, opened 2020-02-10
kubelet-ds: mount iscsiadm and /lib/modules into self-hosted kubelet

invidian:
As this is done by the bootstrap kubelet as well and is required for
running OpenEBS.


invidian:
Just a note, some OpenEBS issue involving CoreOS also suggests mouting `/etc/iscsi` into `kubelet`, so I guess if user has some custom iSCSI settings, kubelet will use them. I'm not sure if we need it for now though. Refs: https://github.com/openebs/openebs/issues/160#issuecomment-315104565

invidian:
Closing in favor of #168.

PR #170, opened 2020-02-11
packet: pass tags to packet_device

mpolednik:
Some applications (such as autoscaler) may require Packet tags to be
set. To avoid having additional tooling to tag the instances we can
pass tags directly to packet_device.

Signed-off-by: Martin Polednik <m.polednik@gmail.com>


alban (approved):
Hi @mpolednik ,

Sorry I hadn't noticed your PR and I had created a similar PR (#179). Your PR looks good to me. I suggest we merge it first if other reviewers agree.

invidian (approved):

rata (approved):
@alban SGTM :-)

PR #177, opened 2020-02-13
Move Packet Calico host protection to Calico helm chart

invidian:
So host protection rules can be updated (as currently protection YAMLs are only copied to controller node, and not to the local assets directory, so if user tries to upgrade the release from files on the disk, it will remove created GNPs, as they are no longer part of `calico` chart).


bootkube/resources/charts/calico/values.yaml
```diff
@@ -9,3 +9,9 @@ calico:
   ipipReadiness: "- --bird-ready"
   podCIDR: 10.2.0.0/16
   networkEncapsulation: "ipipMode: Always"
+  hostEndpoints: []
+  hostProtection:
+    enabled: false
+    nodeCIDR: 0.0.0.0/0
+    managementCIDR:
+    - 0.0.0.0/0
```
rata: This value is also questionable :-D

Can't we force this to be set, instead of have a default value?

aws/flatcar-linux/kubernetes/outputs.tf
```diff
@@ -51,3 +51,24 @@ output "worker_target_group_https" {
   description = "ARN of a target group of workers for HTTPS traffic"
   value       = module.workers.target_group_https
 }
+
+# values.yaml content for all deployed charts.
+output "kubernetes_values" {
```
rata: Don't understand this diff nor the commit message. When, exactly, is this useful?

bootkube/variables.tf
```diff
@@ -139,3 +139,32 @@ variable "external_apiserver_port" {
...
+    labels         = map(string)
+  }))
+
+  default     = []
+}
+
+variable "management_cidrs" {
+  description = "CIDR from where SSH should be allowed on firewall"
+  type        = list(string)
+  default     = ["0.0.0.0/0"]
```
rata: Management_cidrs was mandatory before. Any reason to have this insecure default?

bootkube/variables.tf
```diff
@@ -139,3 +139,32 @@ variable "external_apiserver_port" {
...
+}
+
+variable "node_cidr" {
+  description = "Nodes CIDR for Calico GlobalNetworkPolicies"
+  type        = string
+  default     = "0.0.0.0/0"
```
rata: idem

bootkube/resources/charts/calico/templates/host-endpoint-controller.yaml
```diff
@@ -70,3 +71,4 @@ roleRef:
   kind: ClusterRole
   name: calico-hostendpoint-controller-role
   apiGroup: rbac.authorization.k8s.io
+{{- end }}
```
rata: The controller only works in packet (it has "bond0" hardcoded, etc.). Any reason to move it to the control plane? Just to have it in a general place?

The advantage that I see is that, as it is right now, uses `kubectl` in a simple script and we should keep it in sync with kubernetes version. But not sure about your reasoning and curious about it :)

bootkube/resources/charts/calico/templates/host-protection.yaml
```diff
@@ -0,0 +1,108 @@
...
+spec:
+  selector: nodetype == 'worker'
+  preDNAT: true
+  applyOnForward: true
+  order: 20
+  ingress:
+  - action: Allow
+    protocol: TCP
+    destination:
+      ports: [80, 443]
```
rata: What happens on upgrade if the user has modified this global network policy? (just curious)

bootkube/resources/charts/calico.yaml
```diff
@@ -9,3 +9,27 @@ calico:
   ipipReadiness: "${ipip_readiness}"
   podCIDR: ${pod_cidr}
   networkEncapsulation: "${network_encapsulation}"
+  hostProtection:
+    enabled: ${host_protection}
+    %{~ if node_cidr != "" ~}
+    nodeCIDR: ${node_cidr}
+    %{~ endif ~}
```
rata: `!= ""` is the way to test if the user didn't set it? (new to helm, sorry for the noob question :D)

Can't we force the user to set this (if host_protection is enabled)?

bootkube/resources/charts/calico.yaml
```diff
@@ -9,3 +9,27 @@ calico:
...
+    %{~ if length(management_cidrs) > 0 ~}
+    managementCIDRs:
+    %{~ for cidr in management_cidrs ~}
+    - ${cidr}
+    %{~ endfor ~}
+    %{~ endif ~}
```
rata: If fields are mandatory, hopefully we remove 2 lines here :)

bootkube/resources/charts/calico.yaml
```diff
@@ -9,3 +9,27 @@ calico:
...
+  - name: ${h.name}
+    interfaceName: ${h.interface_name}
+    %{~ if length(h.labels) > 0 ~}
+    labels:
+    %{~ for k, v in h.labels ~}
+      ${k}: ${v}
+    %{~ endfor ~}
+    %{~ endif ~}
+  %{~ endfor ~}
+  %{~ endif ~}
```
rata: Is so horrible and error prone this... :-( I understand we can't do anything about it, but I really dislike it :-)

rata (commented):
@invidian can you please elaborate a little bit on why this is moved to this specific chart?

As I mentioned in the comments, I understand it is to have it all in the same place (even if this makes sense in Packet and **maybe** bare-metal). The advatage I see is that the calico-hostendpoint-controller should be tied to the kubernetes version (it uses `kubectl`, so it shouldn't use kubectl version 1.9 in a 1.17 k8s cluster), but that is only temporal as we should re-write it in go/python and use the regular client libraries (assuming that is not a problem with the client libraries :D).

On one side, I think it might make sense to move it. On the other hand, I'm afraid the charts will be a collection of tons of unrelated things that is very difficult to make sense of. I guess this change is fine, just have this concern for the future maintainability of a sane chart

bootkube/resources/charts/calico.yaml
rata: @invidian will mark this as solved. Just to make a point :joy:

bootkube/resources/charts/calico/values.yaml
invidian: I think I can, but it shouldn't be required when `hostProtection.enabled: false`. I need to check that.

bootkube/variables.tf
invidian: Well, I'm not sure if it has to be mandatory all way down the stack, but I can change it.

bootkube/resources/charts/calico/templates/host-protection.yaml
invidian: The change will be reverted. All changes should be done via `helm`.

bootkube/resources/charts/calico.yaml
invidian: Haha, yeah. I can look if we can use https://www.terraform.io/docs/configuration/functions/yamlencode.html instead of templates.

aws/flatcar-linux/kubernetes/outputs.tf
invidian: This is useful, when you want to programatically get this values from Terraform :) So you don't need to rely on reading file from the disk etc.

invidian: Also, as said in PR description, this code is part of another pull request.

bootkube/resources/charts/calico.yaml
invidian: >!= "" is the way to test if the user didn't set it? (new to helm, sorry for the noob question :D)

I _think_ so, as the structure must be defined in values.yaml, as far as I know.

bootkube/resources/charts/calico/templates/host-endpoint-controller.yaml
invidian: Well, the advantage is, that we maintain YAMLs in a single place. For now it will only be used in Packet, but eventually, I think this controller should be added to Calico Kubernetes controller, so it will still be part of this chart.

That also makes updates easier, as all manifests are handled in the same ways, so we can skip resources in Terraform.

>The advantage that I see is that, as it is right now, uses kubectl in a simple script and we should keep it in sync with kubernetes version. But not sure about your reasoning and curious about it :)

How to we ensure this sync right now?

bootkube/resources/charts/calico.yaml
invidian: Ah, this is actually Terraform template, not helm :)

bootkube/resources/charts/calico/values.yaml
invidian: I removed the defaults and added validation rules.

bootkube/variables.tf
invidian: Done.

bootkube/variables.tf
invidian: Done.

bootkube/variables.tf
invidian: As this is Terraform, this is not possible. I can set it to empty array though, as it must have default value, so it does not complain when installed on platforms other than Packet.

bootkube/resources/charts/calico.yaml
invidian: Hm, I wouldn't actually remove this check, as with Terraform, we are not able to validate, that this value is not empty and having `nodeCIDR: ""` in the result YAML pollutes the file, which may make it harder for debugging.

bootkube/resources/charts/calico.yaml
invidian: Oh, but this function is currently experimental, so I wouldn't actually use it.

iaguis:
I was testing this and it seems the generated GlobalNetworkPolicies don't include the Pod and Service CIDRs now, so pods can't communicate with each other.

johananl:
Following the problem found by Iago (and also regardless), we should verify the host protection policies by checking both "allowed" and "forbidden" packet flows such as the ones described in https://github.com/kinvolk/lokomotive-kubernetes/pull/13#issue-284594707.

bootkube/resources/charts/calico/templates/host-endpoint-controller.yaml
rata: > Well, the advantage is, that we maintain YAMLs in a single place. For now it will only be used in Packet, but eventually, I think this controller should be added to Calico Kubernetes controller, so it will still be part of this chart.

Why? All other platforms (maybe not baremetal, depends on the setup) have firewalls by the cloud provider (security group in aws, etc.). People usually use that, even for kubernetes setups. Why would they use this?

I mean, we can have it. I doubt it will be used outside Packet, that is all.

> How to we ensure this sync right now?

We don't have a process, we need to remember. I think it was a follow-up for some issue to see how to do this.

bootkube/variables.tf
rata: I'd prefer hacks to to affect how other platforms deploy, but the other way around. This way a hack expands to all platforms and all need to be "aware" of this, just because we want that for Packet.

But I don't think we have an easy way out, here :-(

rata: Do you see some more reasonable way?

aws/flatcar-linux/kubernetes/outputs.tf
rata: :ok_hand:

bootkube/resources/charts/calico/templates/host-endpoint-controller.yaml
johananl: The only rationale for introducing GNPs in the first place was lack of provider-level firewalling on Packet. On AWS and similar platforms, using security groups is significantly safer than using GNPs, for multiple reasons. Therefore, I don't see why we would want GNPs on all platforms.

bootkube/resources/charts/calico/templates/host-protection.yaml
rata: :ok_hand:

rata (dismissed):
I think for the reasons explained [here](https://github.com/kinvolk/lokomotive-kubernetes/pull/177#discussion_r379589434) and [here](https://github.com/kinvolk/lokomotive-kubernetes/pull/177#discussion_r379593432) it makes sense to try a totally different approach.

An approach that might make more sense is, when we have a way to install different "components", create such a component and install it for Packet. Then, this chart is clean, the component stays simple... the world is happy :)

Another option is to keep it as it is, and I wouldn't discard it so easily. You basically always want to have this on Packet (or almost always) and there is no way to forget about it like this. When we have a way to install "components", if there is a way to make some mandatory or opt-out, it can make sense too to try something like that.

But I don't think we plan to use this in any other provider, trying to make it generic seems to create some not very nice code/defaults (and is not and not intended to be generic), therefore it's not clear to me it is worth it to be in **this** chart.

What do you think?

invidian:
>I was testing this and it seems the generated GlobalNetworkPolicies don't include the Pod and Service CIDRs now, so pods can't communicate with each other.

Will check that :)

Actually, what I've tried addressing in this PR is the update scenario, where user updates Lokomotive version, re-runs Terraform and then do the update of calico chart using command: `helm upgrade calico -n kube-system <assets dir>/charts/kube-system/calico`.

Currently, with this scenario, GNP **will be removed**, as they are only copied to bootstrap node over SSH and they are not created locally.

I could as well just create local files, but with this proposed approach, I take advantage, that we use Helm for managing manifests and I avoid doing the weird injection of manifests into the Chart, which makes the approach cleaner, as all manifests are managed from a single place. It also moves the manifests templating from Terraform to Helm.

I also tried to put in some improvements to it, so HostEndpoints are now generic and we have slightly better validation rules for the values, which can be extended in the future.

Should I change this PR to just inject local files into assets directory, as we do right now then?

invidian:
Created #180 as an alternative approach.

rata:
> Actually, what I've tried addressing in this PR is the update scenario, where user updates Lokomotive version, re-runs Terraform and then do the update of calico chart using command: `helm upgrade calico -n kube-system <assets dir>/charts/kube-system/calico`.
> 
> Currently, with this scenario, GNP **will be removed**, as they are only copied to bootstrap node over SSH and they are not created locally.

1. Wait, just to be clear: without this PR (i.e. master), GNP rules are removed on upgrades? Wanted to be sure.  Btw, If that is the case, please state that in the PR description next time. Seems way better description of why we care about :-)
1. With this PR, upgrades work fine?
1. What happens if GNP rules are installed with another helm chart? Do upgrades work fine?

Regarding 1, I'm not familiar with helm still to know, but why is that? The helm chart doesn't "own" just the things it installs? This surprises me a little, and seems like it can create TONs of problems. For example, if the user uses [kubernetes network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) won't they break on upgrades? Or if the user creates a new GNP rule, won't they break?

I really hope I'm missing something and those things won't happen :-D

invidian:
> Btw, If that is the case, please state that in the PR description next time. Seems way better description of why we care about :-)

Yes, I should've put better description :)

>Wait, just to be clear: without this PR (i.e. master), GNP rules are removed on upgrades? Wanted to be sure.

If you do `helm upgrade -n kube-system calico <assets dir>/charts/kube-system/calico` right now, then yes, GNP will be removed. This is not a documented way of upgrading, but this would be reasonable one for experienced users.

>With this PR, upgrades work fine?

Yes, with this PR, all charts related files are always rendered to local directory and then optional features (host protection) is driven via Helm templating, so yes, updates (as described in the process above) works fine with this PR.

>What happens if GNP rules are installed with another helm chart? Do upgrades work fine?

Yes, it should be fine to install them as a separated chart. We just need to make sure they are being installed after Calico itself (to have CRDs registered). I'm actually working on it right now, I hope that still make sense.

>Regarding 1, I'm not familiar with helm still to know, but why is that? The helm chart doesn't "own" just the things it installs? This surprises me a little, and seems like it can create TONs of problems. For example, if the user uses kubernetes network policies won't they break on upgrades? Or if the user creates a new GNP rule, won't they break?

The reason for that is that protection YAMLs are appended to bootkube assets on controller node, as part of bootstrapping. Those file **won't** be created on local disk, so when running `helm upgrade`, Helm detects that the manifests are not there anymore and removes them from the cluster, as it tracks the orphans.

rata:
> > Wait, just to be clear: without this PR (i.e. master), GNP rules are removed on upgrades? Wanted to be sure.
> 
> If you do `helm upgrade -n kube-system calico <assets dir>/charts/kube-system/calico` right now, then yes, GNP will be removed. This is not a documented way of upgrading, but this would be reasonable one for experienced users.

But if you upgrade in a documented way, does it break too? (don't have a cluster handy to test now :))

> > With this PR, upgrades work fine?
> 
> Yes, with this PR, all charts related files are always rendered to local directory and then optional features (host protection) is driven via Helm templating, so yes, updates (as described in the process above) works fine with this PR.

:ok_hand: thanks!

> 
> > What happens if GNP rules are installed with another helm chart? Do upgrades work fine?
> 
> Yes, it should be fine to install them as a separated chart. We just need to make sure they are being installed after Calico itself (to have CRDs registered). I'm actually working on it right now, I hope that still make sense.

I think it makes sense, don't know what others think :-). What do you think? :)

> > Regarding 1, I'm not familiar with helm still to know, but why is that? The helm chart doesn't "own" just the things it installs? This surprises me a little, and seems like it can create TONs of problems. For example, if the user uses kubernetes network policies won't they break on upgrades? Or if the user creates a new GNP rule, won't they break?
> 
> The reason for that is that protection YAMLs are appended to bootkube assets on controller node, as part of bootstrapping. Those file **won't** be created on local disk, so when running `helm upgrade`, Helm detects that the manifests are not there anymore and removes them from the cluster, as it tracks the orphans.

Hmm, okay. I don't follow all, but seems reasonable

(is something like this more or less true? Just to see if I'm following: helm tracks them initially as they are copied to the manifests folder and installs them, but, a they are not part of the chart, when upgrading those files are not generated and therefore considers them orphans and deletes them?)

invidian:
>(is something like this more or less true? Just to see if I'm following: helm tracks them initially as they are copied to the manifests folder and installs them, but, a they are not part of the chart, when upgrading those files are not generated and therefore considers them orphans and deletes them?)

Correct.

>But if you upgrade in a documented way, does it break too? (don't have a cluster handy to test now :))

Hm, now that I think about it, we no longer have plain YAML files rendered once we moved to Helm, so documented way of updating is out of date right now :/

rata:
> > (is something like this more or less true? Just to see if I'm following: helm tracks them initially as they are copied to the manifests folder and installs them, but, a they are not part of the chart, when upgrading those files are not generated and therefore considers them orphans and deletes them?)
> 
> Correct.

Thanks!

> 
> > But if you upgrade in a documented way, does it break too? (don't have a cluster handy to test now :))
> 
> Hm, now that I think about it, we no longer have plain YAML files rendered once we moved to Helm, so documented way of updating is out of date right now :/

Right, but I meant "whatever is the documented way".

In any case, want to create an issue to update the docs later? :)

invidian:
Updated the PR with separated helm chart for host protection.

iaguis (approved):
lgtm

PR #180, opened 2020-02-17
packet: copy Calico manifests to local assets rather than over SSH

invidian:
So if user wants to roll an update of Calio using 'helm upgrade' using
local assets, it won't remove those manifests.

As a simpler alternative to #177.


iaguis (commented):
LGTM, but I want to test it before approving it. I'll do that now

invidian:
Oh, it seems that won't work, as `template_dir` we use to copy Calico chart will always overwrite the destination directory, so it will be causing the diff all the time. Also, I've tried adding dependencies on it, but that doesn't seem to work very well. I guess next attempt will be to modify #177 and install host protection as separated chart then.

PR #182, opened 2020-02-18
packet/flatcar-linux/kubernetes: add etcd records at the end

### LINKED ISSUE #181: Adding more controlplane nodes fails on Packet
With following error:
```
module.dns.aws_route53_record.dns-records[2]: Creation complete after 41s [id=Z1QEP94KB3YNR3_mateusz.example.com._A]
module.dns.aws_route53_record.dns-records[1]: Creation complete after 41s [id=Z1QEP94KB3YNR3_mateusz-etcd1.example.com._A]

Error: [ERR]: Error building changeset: InvalidChangeBatch: [Tried to create resource record set [name='mateusz-private.dev.example.com.', type='A'] but it already exists]
	status code: 400, request id: fee05404-59ba-485b-8d0d-bb755ec683de

  on .terraform/modules/dns/main.tf line 17, in resource "aws_route53_record" "dns-records":
  17: resource "aws_route53_record" "dns-records" {



```

CC @mauriciovasquezbernal

invidian:
Otherwise, when adding controller node, all records are shifted by one,
which triggers re-creating them, which may cause the downtime and
currently produces conflicting records, which breaks adding controller
nodes.

With this commit, etcd records will be added at the end of the
collection and first two entries should always remain the same and be
updated in place.

Closes #181

### DIFF
--- a/packet/flatcar-linux/kubernetes/controllers.tf
+++ b/packet/flatcar-linux/kubernetes/controllers.tf
@@ -4,16 +4,6 @@ locals {
   etcd_fqdn         = [for index, device in packet_device.controllers : format("%s-etcd%d.%s.", var.cluster_name, index, var.dns_zone)]
 
   dns_entries = concat(
-    # etcd
-    [
-      for index, device in packet_device.controllers :
-      {
-        name    = local.etcd_fqdn[index],
-        type    = "A",
-        ttl     = 300,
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
I haven't tested it myself, tests are passing and I'm confident it would not make things worse.

iaguis (approved):

PR #183, opened 2020-02-18
Add BGP-related node labels

johananl:
When BGP is enabled for a worker pool, create node labels to be picked up by the "node peers" MetalLB feature.

https://github.com/metallb/metallb/issues/529

### DIFF
--- a/packet/flatcar-linux/kubernetes/workers/workers.tf
+++ b/packet/flatcar-linux/kubernetes/workers/workers.tf
@@ -44,26 +44,35 @@ resource "packet_bgp_session" "bgp" {
   address_family = "ipv4"
 }
 
+# BGP node labels.
+locals {
+  my_asn = format("metallb.universe.tf/my-asn=%d", data.packet_project.project.bgp_config.0.asn)
+  # Packet always uses ASN 65530 as the remote ASN for local BGP.
+  peer_asn = format("metallb.universe.tf/peer-asn=%d", 65530)
+}
+
 data "ct_config" "ignitions" {
-  content  = data.template_file.configs.rendered
+  content = templatefile(
+    "${path.module}/cl/worker.yaml.tmpl",
+    {
+      os_arch               = var.os_arch
+      kubeconfig            = indent(10, var.kubeconfig)
+      ssh_keys              = jsonencode(var.ssh_keys)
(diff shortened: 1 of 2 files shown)


packet/flatcar-linux/kubernetes/workers/workers.tf
```diff
@@ -44,26 +44,35 @@ resource "packet_bgp_session" "bgp" {
...
+locals {
+  my_asn = format("metallb.universe.tf/my-asn=%d", data.packet_project.project.bgp_config.0.asn)
+  # Packet always uses ASN 65530 as the remote ASN for local BGP.
+  peer_asn = format("metallb.universe.tf/peer-asn=%d", 65530)
+}
+
 data "ct_config" "ignitions" {
-  content  = data.template_file.configs.rendered
+  content = templatefile(
+    "${path.module}/cl/worker.yaml.tmpl",
```
invidian, line 56: Do we need newlines here? What about `templatefile("${path.module}/cl/worker.yaml.tmpl", {` syntax? Is `terraform fmt` opinionated about it?

invidian (commented):
Just one nit, otherwise looks good.

iaguis (dismissed):
lgtm

packet/flatcar-linux/kubernetes/workers/workers.tf
johananl: >Do we need newlines here?

We can't fit the map in one line so we need at least *some* newlines. Regarding having the first argument to `templatefile()` on its own line, I personally find it more readable to have one argument per line than having the first argument on the same line with the function call and have newlines starting from the 2nd argument.

Looking at other similar instances, this seems consistent with existing code to me. Examples:

https://github.com/kinvolk/lokomotive-kubernetes/blob/b4d5a94db8662be29182b54c678224fa51461835/packet/flatcar-linux/kubernetes/workers/workers.tf#L14
https://github.com/kinvolk/lokomotive-kubernetes/blob/b4d5a94db8662be29182b54c678224fa51461835/packet/flatcar-linux/kubernetes/ssh.tf#L154

The file is `terraform fmt`-ed.

invidian (approved):
:see_no_evil:

iaguis (approved):

